PR #6, opened 2020-10-01
[DO NOT MERGE] Mern/ingestion sdk

sebastianwoinar:
## Types of changes

## Checklist:

- [ ] My code follows the code style of this project.


sdk-core/src/main/java/care/data4life/sdk/ApiService.java
```diff
@@ -245,12 +293,37 @@ Completable deleteRecord(String alias, String recordId, String userId) {
                 .subscribeOn(Schedulers.io());
     }
 
+    /**
+     * Carry out needed logout actions.
+     * <p>
+     * When using refresh token, this will revoke the OAuth access.
+     * When using a static access token, nothing will be done.
```
sebastianwoinar, line 300: Why is that not possible?
From what I see, only the access token is used for ending the user's session in vega.

sdk-ingestion/src/main/java/care/data4life/sdk/AccessTokenDecoder.kt
```diff
@@ -0,0 +1,73 @@
...
+    private val moshi: Moshi = Moshi.Builder().build()
+
+    /**
+     * Extract the client ID from the OAuth access token.
+     *
+     * @param accessToken Full access token
+     *
+     * @return Client ID
+     */
+    fun extractClientId(accessToken: ByteArray?): String {
```
sebastianwoinar, line 49: 👍🏻

sdk-ingestion/src/main/java/care/data4life/sdk/AccessTokenDecoder.kt
```diff
@@ -0,0 +1,73 @@
...
+            if (null == clientId) {
+                throw Exception("Did mot find client ID in JWT")
+            }
+        } catch (e: Exception) {
+            throw D4LRuntimeException("Access token not correctly formatted")
+        }
+        return clientId
+    }
+
+}
```
sebastianwoinar, line 73: 😱
Please add a newline.
```suggestion
}

```

sdk-ingestion/src/main/java/care/data4life/sdk/D4LJwtPayload.kt
```diff
@@ -0,0 +1,59 @@
...
+        val exp: Double? = null,                            // "Expiration Time" claim
+        val nbf: Double? = null,                            // "Not Before" claim
+        val iat: Double? = null,                            // "Issued At" claim
+        val jti: String? = null,                            // "JWT ID" claim
+        /* Private D4L claims */
+        @field:Json("ghc:scope") val ghc_scope: String? = null,      // Scopes - could consider adding a type adapter for the scopes that parses them in to a list
+        @field:Json("ghc:uid") val ghc_uid: String? = null,          // The source User ID which requested the JWT (not always the subject)
+        @field:Json("ghc:cid") val ghc_cid: String? = null,          // The client ID which requested the JWT
+        @field:Json("ghc:aid") val ghc_aid: String? = null           // The app ID which requested the JWT
+)
```
sebastianwoinar, line 59:
```suggestion
)

```

sebastianwoinar (commented):
Looks already very good

sdk-core/src/main/java/care/data4life/sdk/ApiService.java
sebastianwoinar: It is the refresh token

sdk-ingestion/src/main/java/care/data4life/sdk/Data4LifeClient.java
```diff
@@ -0,0 +1,145 @@
...
+ * Any usage of the SDK for other purposes, including usage for the development of
+ * applications/third-party applications shall require the conclusion of a license agreement
+ * between you and D4L.
+ *
+ * If you are interested in licensing the SDK for your own applications/third-party
+ * applications and/or if you’d like to contribute to the development of the SDK, please
+ * contact D4L by email to help@data4life.care.
+ */
+
+package care.data4life.sdk;
```
sebastianwoinar, line 17: Why is it in this package?
This create a clash with the other SDKs.

For instance the test-broker is using the Data4LifeClient from the jvm-core and the ingestion-sdk at the same time.

sebastianwoinar: @salehgamil

sebastianwoinar: I feel, that having the same class (`care.data4life.sdk.Data4LifeClient`) with different signatures in different jars is not very good practice, is it?

How about moving this into `care.data4life.ingestion-sdk` ??

me-d4l: I believe keeping the name was actually a request from Wolf. Currently, the sdk-android and sdk-jvm packages also both define classes called `Data4LifeClient` in the `care.data4life.sdk` package, and they also have different signatures from the constructors & the `init()` factory methods. So it seems the idea is to have the package be the same and just binding in the right JAR. But for my sake, we could arrange it differently.

sebastianwoinar: @aps do you have some background on this?
For us it would be really benefitial to have the ingestion SDK's Data4LifeClient in a different package.

aps: Yes as @me-d4l is saying, since the project is meant to be a multiplatform project the android and jvm classes will have the same package. 

I don't know what the purpose of the `ingestion-sdk` is, but if it's seen as a new platform then the package should stay the same for the consistency reasons. 

> For instance the test-broker is using the Data4LifeClient from the jvm-core and the ingestion-sdk at the same time.

Why do you want to have both of them in there?

sebastianwoinar: It's only for testing purposes. 
The ingestion SDK has only limited capabilities (append only) and with the JVM SDK we are testing if the uploaded resources are there.
Right now, in the very same spring application are the two SDKs reachable on different endpoints.
Of course we could extract that and put this into different spring applications but it is a but cumbersome.

But well, if this is the way you want to have it, than @salehgamil can look into the changes for the test-broker next week. 😬

aps: I just double checked, but it actually cannot get moved to another package. All the core components are in the `.sdk` package which won't be accessible from the `.integration-sdk` package due to the visibility being package-private.

sebastianwoinar: Ok, thanks for looking into this. Then we must change it on our site.
🙏🏻

sdk-core/src/main/java/care/data4life/sdk/ApiService.java
```diff
@@ -79,15 +80,29 @@
     private OkHttpClient client;
     private final String clientName;
     private final boolean debug;
-
-
+    private final boolean useStaticToken;
+
+    /**
```
wmontwe, line 85: Nice 👍

sdk-core/src/main/java/care/data4life/sdk/ApiService.java
```diff
@@ -79,15 +80,29 @@
...
+     * Full constructor.
+     *
+     * @param oAuthService        OAuth service
+     * @param environment         Deployment environment
+     * @param clientID            Client ID
+     * @param clientSecret        Client secret
+     * @param platform            Usage platform (D4L, S4H)
+     * @param connectivityService Connectivity service
+     * @param clientName          Client name
+     * @param useStaticToken      FLag indicating if a static OAuth token should be used (no token reneval)
```
wmontwe, line 95: typo -> renewal

sdk-core/src/main/java/care/data4life/sdk/ApiService.java
```diff
@@ -245,12 +293,37 @@ Completable deleteRecord(String alias, String recordId, String userId) {
...
+        if (this.useStaticToken) {
+            throw new D4LRuntimeException("Cannot log out when using a static access token!");
+        }
         return Single
                 .fromCallable(() -> oAuthService.getRefreshToken(alias))
                 .flatMapCompletable(token -> service.logout(alias, token));
+
     }
 
+    /**
```
wmontwe, line 315: it's quite unusual to have a documentation on private class members

sdk-core/src/main/java/care/data4life/sdk/ApiService.java
```diff
@@ -289,6 +362,35 @@ private Response intercept(Interceptor.Chain chain) throws IOException {
         }
         return chain.proceed(request);
     }
+
+    /**
```
wmontwe, line 366: same as above private docs

sdk-ingestion/src/main/java/care/data4life/sdk/Data4LifeClient.java
wmontwe: The library is not designed to be used in different flavors alongside in the same project.  But an  idea: you could have a `Data4LifeIngestionClient` that implements `BaseClient` to avoid collisions in the `.sdk`package.

wmontwe (changes requested):

sdk-core/src/main/java/care/data4life/sdk/ApiService.java
me-d4l: Thank, fixed!

sdk-core/src/main/java/care/data4life/sdk/ApiService.java
me-d4l: Right, this would be more for internal documentation purposes, I wrote it up to capture my understanding of the function. Would you prefer to have it removed? (e.g. for consistency or if it messes up automatic generation of documentation by including private classes or for consistency)

me-d4l:
Closing this as it was superseded by #12 (identical code, but different branch name)

me-d4l:
Re-opening as the open comments otherwise do not show in new PR.

PR #13, opened 2020-11-09
[NAT-124] Add arbitrary appdata support

ATizik:
## Description
Adds support for storing arbitrary data

## Motivation and Context

Initially needed for DonorKey from Data Donation feature, but can be used for more use cases

Additional set of CRUD methods are implemented, and annotation support is added on top of tags implementation to differentiate between different types of arbitrary data. Essentially an annotation is just a tag with key "custom" and some consumer provided value, which should be used to differentiate between types. We also expose `AppDataRecord` as an analogous class to the usual `Record`, which just stores byte array of arbitrary consumer-provided data. 

## How Has This Been Tested?

Several tests were added, and parts of the implementation are already covered by tests because a lot of underlying code is reused from records. Sample also implements a simple check of upload-download appData to check that the contents and annotations match.
Manual crossplatform test with iOS-uploaded Records was done to check annotations are readable on both platforms.

Annotations are implemented as an additional parameter on new methods, and old methods are left intact for compatibility. Invoking old methods are analogous to providing empty annotation list. 

## Types of changes

## Checklist:

- [x] I have updated the documentation accordingly.


wmontwe:
Why are there 0 tests?

sdk-core/src/main/java/care/data4life/sdk/model/DonorKey.kt
```diff
@@ -0,0 +1,51 @@
...
+import com.squareup.moshi.JsonReader
+import com.squareup.moshi.JsonWriter
+import com.squareup.moshi.Moshi
+import com.squareup.moshi.ToJson
+import okio.Buffer
+import okio.ByteString
+import java.io.CharArrayWriter
+import java.lang.reflect.Type
+
+data class DonorKey(
```
wmontwe, line 31: This should not be part of the SDK

ATizik:
There is one integration test in sample project, but since the final implementation wasn't settled until now I didn't write documentation/more tests

sdk-core/src/main/java/care/data4life/sdk/AppDataResource.kt
```diff
@@ -0,0 +1,49 @@
...
+package care.data4life.sdk
+
+import care.data4life.crypto.GCKey
+import care.data4life.sdk.model.Meta
+
+/**
+ * AppDataRecord is used to store arbitrary data, analogous to Record
+ * @see care.data4life.sdk.model.Record
+ */
+data class AppDataRecord(
```
bitPogo, line 26: If it is analogous to Record, why they do not share a common interface?

sdk-core/src/main/java/care/data4life/sdk/CryptoService.java
```diff
@@ -147,6 +150,27 @@ public CryptoService(
...
+    public Single<char[]> decryptToCharArray(GCKey key, String dataBase64) {
+        return Single
+                .fromCallable(() -> base64.decode(dataBase64))
+                .flatMap(decoded -> decrypt(key, decoded))
+                .map(decrypted -> StandardCharsets.UTF_8.decode(ByteBuffer.wrap(decrypted)).array())
+                .onErrorResumeNext(error -> {
+                    error(error, "Failed to decrypt string");
+                    return Single.error((D4LException) new CryptoException.DecryptionFailed("Failed to decrypt string"));
+                });
+    }
```
bitPogo, line 172: Both could be merged with the already existing string methods, to reduce code

bitPogo (commented):
Besides the obvious.

sdk-core/src/main/java/care/data4life/sdk/model/Record.java
```diff
@@ -16,15 +16,17 @@
...
 
+import java.util.List;
 import java.util.Objects;
 
 import care.data4life.fhir.stu3.model.DomainResource;
 
 public class Record<T extends DomainResource> {
     private T fhirResource;
     private Meta meta;
+    private List<String> annotations;
```
bitPogo, line 27: Is this a leftover?

wmontwe:
this is redone in #22

PR #15, opened 2020-11-16
[SDK-407] Add config versioning

ouafae31:
## Description

SDK versioning support [Ticket_407](https://gesundheitscloud.atlassian.net/browse/SDK-407?atlOrigin=eyJpIjoiMzlhYzQzMDJiYjdjNGY5Yjg4ZTMyOWQ4MzQ0YzA4ZGUiLCJwIjoiaiJ9)
## Motivation and Context

This PR depends on the PR in the config repo https://github.com/gesundheitscloud/hc-sdk-config/pull/12
## How Has This Been Tested?

Supported = last supported version
Deprecated = not the last version but still supported, the user should update the SDK(no warning was implemented for now)
Unsupported = not the last version and not supported anymore, the user can not use the SDK{we throw an error)
To stop supporting a version, change the status to "unsupported".
If the version is not  found, the default value is 'supported'
These changes were not added to the Java project.

## Types of changes

- [X] New feature (non-breaking change which adds functionality)

## Checklist:

- [ ] My code follows the code style of this project.
- [X] I have updated the changelog accordingly.
- [X] I have added tests to cover my changes.
- [X] All new and existing tests passed.


gradle.properties
```diff
@@ -40,4 +40,4 @@ org.gradle.daemon=false
 org.gradle.jvmargs=-Xmx1536m
 
 gpr.user=ouafae31
-gpr.key=8ffa3a646b94dd9ae1eb6cf14eda1887ca8ce367
```
wmontwe: you need to rotate your key please

sdk-core/src/test/java/care/data4life/sdk/auth/UserServiceTest.kt
```diff
@@ -0,0 +1,251 @@
...
+        apiService = mockk<ApiService>()
+        cryptoService = mockk()
+        userService = spyk(UserService(USER_ALIAS, authService, apiService, storage, cryptoService))
+    }
+
+    @Test
+    fun `isLoggedIn_shouldReturnTrue`() {
+        //given
+        every { authService.isAuthorized(USER_ALIAS) } returns true
+        every { storage.getSecret(AUTH_STATE) } answers {
```
bitPogo, line 74: `returns` without the braces (we do not need a closure here)

sdk-core/src/test/java/care/data4life/sdk/auth/UserServiceTest.kt
```diff
@@ -0,0 +1,251 @@
...
+            "auth_state".toCharArray()
+        }
+        val mockGCKey = mockk<GCKey>()
+        every { mockGCKey.getSymmetricKey() } returns mockk<GCSymmetricKey>()
```
bitPogo, line 78: You do not need to add the type here on the mockk, it can derive it automatically.

sdk-core/src/test/java/care/data4life/sdk/auth/UserServiceTest.kt
```diff
@@ -0,0 +1,251 @@
...
+        // then
+        testSubscriber
+            .assertNoErrors()
+            .assertValue(true)
+    }
+
+    @Test
+    fun `isLoggedIn_shouldReturnFalse`() {
+        //given
+        every { storage.getSecret(AUTH_STATE) } answers {
```
bitPogo, line 94: `returns` without the braces (we do not need a closure here)

sdk-core/src/test/java/care/data4life/sdk/auth/UserServiceTest.kt
```diff
@@ -0,0 +1,251 @@
...
+            "auth_state".toCharArray()
+        }
+        val mockGCKey = mockk<GCKey>()
+        every { mockGCKey.getSymmetricKey() } returns mockk<GCSymmetricKey>()
```
bitPogo, line 98: You do not need to add the type here on the mockk, it can derive it automatically.

sdk-core/src/test/java/care/data4life/sdk/auth/UserServiceTest.kt
```diff
@@ -0,0 +1,251 @@
...
+        every { authService.refreshAccessToken(USER_ALIAS) } returns mockk<String>()
+        // when
+        val testSubscriber = userService.getSessionToken(USER_ALIAS)
+            .test()
+            .await()
+
+        // then
+        testSubscriber
+            .assertNoValues()
+    }
```
bitPogo, line 138: I guess this can go then, since the authService does not allows nulled Strings as return value any longer.

sdk-core/src/test/java/care/data4life/sdk/auth/UserServiceTest.kt
```diff
@@ -0,0 +1,251 @@
...
+        // when
+        val testSubscriber = userService.logout()
+            .test()
+            .await()
+
+        // then
+        testSubscriber
+            .assertError(Exception::class.java)
+            .assertNotComplete()
+        verify { storage wasNot Called }
```
bitPogo, line 202: That's nice! A good pattern to follow.

sdk-core/src/test/java/care/data4life/sdk/auth/UserServiceTest.kt
```diff
@@ -0,0 +1,251 @@
...
+            .test()
+            .await()
+
+        // then
+        testSubscriber
+            .assertNoValues()
+    }
+
+    @Test
+    fun `Given VersionList is fetched, when the call fails return true`() {
```
bitPogo, line 141: Also I'd like to stay with the pattern, we already introduced, but this just a matter of taste.

sdk-core/src/test/java/care/data4life/sdk/auth/UserServiceTest.kt
```diff
@@ -0,0 +1,251 @@
...
+        val testSubscriber = userService.getVersionInfo("1.9.0")
+            ?.test()
+            ?.await()
+        //then
+        testSubscriber?.assertValue(true)
+        verify(exactly = 1) { Log.error(error, "Version not supported") }
+    }
+
+    @Test
+    fun `Given VersionList is fetched, when version supported return true`() {
```
bitPogo, line 157: Also I'd like to stay with the pattern, we already introduced, but this just a matter of taste.

sdk-core/src/test/java/care/data4life/sdk/auth/UserServiceTest.kt
```diff
@@ -0,0 +1,251 @@
...
+        val error = RuntimeException("error")
+        val response: Single<VersionList> = Single.fromCallable { throw error }
+        mockkObject(Log)
+        every { apiService.fetchVersionInfo() } returns response
+        //when
+        val testSubscriber = userService.getVersionInfo("1.9.0")
+            ?.test()
+            ?.await()
+        //then
+        testSubscriber?.assertValue(true)
```
bitPogo, line 152: I guess the null checks should not be necessary.

sdk-core/src/test/java/care/data4life/sdk/auth/UserServiceTest.kt
```diff
@@ -0,0 +1,251 @@
...
+        val versions: VersionList = mockk()
+        val response: Single<VersionList> = Single.just(versions)
+        val currentVersion = "1.9.0"
+        every { apiService.fetchVersionInfo() } returns response
+        every { versions.isSupported(currentVersion) } returns true
+        //when
+        val testSubscriber = userService.getVersionInfo("1.9.0")
+            ?.test()
+        //then
+        testSubscriber?.assertValue(true)
```
bitPogo, line 167: I guess the null checks should not be necessary.

sdk-core/src/test/java/care/data4life/sdk/auth/UserServiceTest.kt
```diff
@@ -0,0 +1,251 @@
...
+
+    @Test
+    fun `Given VersionList is fetched, when version supported return true`() {
+        val versions: VersionList = mockk()
+        val response: Single<VersionList> = Single.just(versions)
+        val currentVersion = "1.9.0"
+        every { apiService.fetchVersionInfo() } returns response
+        every { versions.isSupported(currentVersion) } returns true
+        //when
+        val testSubscriber = userService.getVersionInfo("1.9.0")
```
bitPogo, line 164: Possibility you can reuse `currentVersion` here.

sdk-core/src/test/java/care/data4life/sdk/auth/UserServiceTest.kt
```diff
@@ -0,0 +1,251 @@
...
+
+    @Test
+    fun `Given VersionList is fetched, when the call fails return true`() {
+        //given
+        val error = RuntimeException("error")
+        val response: Single<VersionList> = Single.fromCallable { throw error }
+        mockkObject(Log)
+        every { apiService.fetchVersionInfo() } returns response
+        //when
+        val testSubscriber = userService.getVersionInfo("1.9.0")
```
bitPogo, line 148: Possibility you can reuse `currentVersion` here.

sdk-core/src/test/java/care/data4life/sdk/auth/UserServiceTest.kt
```diff
@@ -0,0 +1,251 @@
...
+        testSubscriber?.assertValue(true)
+            ?.assertNoErrors()
+
+    }
+
+    @Test
+    fun `logout should Return Success`() {
+        //given
+        every { apiService.logout(USER_ALIAS) } returns Completable.complete()
+        every { storage.clear() } just Runs
```
bitPogo, line 176: Oh `just Runs`, looks nice! Good pattern to follow!

sdk-core/src/test/java/care/data4life/sdk/auth/UserServiceTest.kt
```diff
@@ -0,0 +1,251 @@
...
+        // when
+        val testSubscriber = userService.logout()
+            .test()
+            .await()
+
+        // then
+        testSubscriber
+            .assertNoErrors()
+            .assertComplete()
+        verify { storage.clear() }
```
bitPogo, line 186: Maybe `verify(exactly = 1) { storage.clear() }

sdk-core/src/test/java/care/data4life/sdk/auth/UserServiceTest.kt
```diff
@@ -0,0 +1,251 @@
...
+    private lateinit var userService: UserService
+    private lateinit var cryptoService: CryptoService
+    private lateinit var authService: AuthorizationService
+
+    @Before
+    fun setUp() {
+        authService = mockk<AuthorizationService>()
+        userService = mockk<UserService>()
+        storage = mockk<CryptoSecureStore>()
+        apiService = mockk<ApiService>()
```
bitPogo, line 65: You do not need to add the type here on the mockk, it can derive it automatically.

sdk-core/src/test/java/care/data4life/sdk/auth/UserServiceTest.kt
```diff
@@ -0,0 +1,251 @@
...
+            cryptoService.symDecryptSymmetricKey(any(), any())
+        } returns Single.just(mockk<GCKey>())
+        // when
+        val testSubscriber = userService.finishLogin(true)
+            .test()
+
+
+        // then
+        testSubscriber
+            .assertValue { aBoolean: Boolean? -> aBoolean!! }
```
bitPogo, line 232: You can replace `aBoolean: Boolean? -> aBoolean!!` simply by `it` or write `aBoolean -> aBoolean` to get rid of the unnecessary typing.

sdk-core/src/test/java/care/data4life/sdk/auth/UserServiceTest.kt
```diff
@@ -0,0 +1,251 @@
...
+        testSubscriber
+            .assertError(Exception::class.java)
+            .assertNotComplete()
+        verify { storage wasNot Called }
+    }
+
+    @Test
+    fun `finish Login should Return Boolean Success`() {
+        //given
+        val userInfo = mockk<UserInfo>()
```
bitPogo, line 208: Could you please make here a usage of variables? So this can get rid of the `any()` parameter?

bitPogo (changes requested):
There actually some things, which needs to be fixed.

bitPogo:
And you need to update the Changelog.

sdk-core/src/test/java/care/data4life/sdk/auth/UserServiceTest.kt
ouafae31: why is it important to not use any? in this test I only want to check finishLogin method.

sdk-core/src/main/java/care/data4life/sdk/auth/UserService.kt
```diff
@@ -75,4 +76,12 @@ class UserService(
         return Single.fromCallable { authService.refreshAccessToken(alias) }
     }
 
+    fun getVersionInfo(currentVersion: String): Single<Boolean>? {
```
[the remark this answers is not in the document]
bitPogo: @ouafae31 This still needs to be changed.

sdk-core/src/main/java/care/data4life/sdk/network/model/VersionList.kt
```diff
@@ -0,0 +1,36 @@
...
+data class VersionList(
+        @field:Json(name = "versions")
+        override val versions: List<Version>
+) : NetworkModelContract.VersionList {
+
+
+    fun isSupported(currentVersion: String): Boolean {
+        var isSupported = true
+        for (version in versions) {
+            if (version.name == currentVersion.substring(0,5) && version.status == NetworkModelContract.Version.KEY_UNSUPPORTED) {
```
bitPogo, line 29: What is that supposed to do?

sdk-core/src/test/java/care/data4life/sdk/auth/UserServiceTest.kt
bitPogo: Much better now, thanks!

bitPogo (changes requested):
One change still needs to be done and a open question.

sdk-core/src/main/java/care/data4life/sdk/network/model/VersionList.kt
ouafae31: getVersionInfo should be changed to what?

bitPogo: No the `substring(x,y)` part. (sry, I forget always does not mark single parts in a line)

ouafae31: If the version exit and has status UNSUPPORTED: then we don't support this version and must fails the login.

bitPogo: True, but why we need the a substring of the version for that? (That's the bit I do not get)
So maybe make an example, which that should cover.

ouafae31: I added it to run under debug as well. Because the build name = 1.9.0-config-debug-something so this version will not be found unless you trim it to 1.9.0

bitPogo: To be honest, I am a bit uncertain about this, for some reasons. However if you wanna cover such case I guess you should not use substring since it has a fixed range. So for example `1.9.32` would be cut to `1.9.3`, which it should not. You can use a regex for that or split `-` with the assumption that it will be the only delimiter we ever use.

bitPogo: Also it needs to be covered by a test as well.

ouafae31: The test is added already. I thought about that but I don't think we support 1.9.32..the third digits is for fixing bugs, the second for features.

bitPogo: I mean, I can make another example - the upcoming `1.10.0`, which will have a similar issue. Also I cannot spot any test, which tests this trimming behaviour. Did you pushed the test?

ouafae31: not a specific test just set the example lijke the debug version.

sdk-core/src/main/java/care/data4life/sdk/auth/UserService.kt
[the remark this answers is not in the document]
bitPogo: Maybe you overlooked that one..

sdk-core/src/main/java/care/data4life/sdk/network/model/VersionList.kt
```diff
@@ -0,0 +1,38 @@
...
+
+data class VersionList(
+        @field:Json(name = "versions")
+        override val versions: List<Version>
+) : NetworkModelContract.VersionList {
+
+
+    fun isSupported(currentVersion: String): Boolean {
+        var isSupported = true
+        val regex = Regex("^(\\d+\\.)?(\\d+\\.)?(\\d)")
```
bitPogo, line 28: Nitpick: are those `?` needed (aka - is there a need to have the digits  optional?)

bitPogo (commented):
Thank you for the changes, there are minor issues left - please resolve then, so we can finally merge this PR.

bitPogo (approved):
LGTM

bitPogo:
I guess your branch names is ill named - it needs to be `feature/SDK-407/add-config-versioning`.
You can find a guide for renaming remote branches [here](https://multiplestates.wordpress.com/2015/02/05/rename-a-local-and-remote-branch-in-git/).

PR #26, opened 2020-12-16
[SDK-230] Fix string in crypto service

ouafae31:
For security reasons, we want to avoid using String keys.
## Description

Change String keys to CharArray in Java and Android

I changed all keys types from String to CharArray, except 
interface Key {

    fun getAlgorithm(): String

    fun getFormat(): String

    fun getEncoded(): ByteArray
}
As in CryptoService this gets converted to GCkey anyway in this function: 
convertPrivateKeyPemStringToGCKeyPair.

## How Has This Been Tested?

The build has some issues in RecordService so I could not run the tests to be sure nothing was broken.
## Types of changes

- [X] Bug fix (non-breaking change which fixes an issue)

## Checklist:

- [X] My code follows the code style of this project.


me-d4l (commented):
Thanks for pinging us on this, we will consider whether we should change on PEM key interface in this direction as well. The change to the file I added look fine, for the rests I cannot tell.

wmontwe:
Tests will be extracted to another pull-request and we address this issue after mooving out securestore and crypto from this repository.

PR #39, opened 2021-01-13
Fix checkRestrictions

bitPogo:
## Description
This addresses the issue, which had been experienced with the [Patient testset](https://github.com/gesundheitscloud/s4h-fhir-profiles/blob/94c4afe43e60bf5c8c4faa5c67276d9ce533789f/examples/s4h-patient-example.patient.json#L1).

## Motivation and Context
The current implementation for `checkDataRestrictions` breaks if a Resource contains Attachments (or at least seen by the FhirAttachmentHelper as such), which are not extractible out of the Resource.

The Attachements are not longer referenced as List, there are now nullable. Additionally after the extraction of the Attachments is now a null check and if the Attachments are null `checkDataRestrictions` simply returns.
Also to avoid any confusion the style of `_uploadData` and `checkDataRestrictions` had been aligned.

## How Has This Been Tested?
A new Unittest had been added, which makes use of the [Patient testset](https://github.com/gesundheitscloud/s4h-fhir-profiles/blob/94c4afe43e60bf5c8c4faa5c67276d9ce533789f/examples/s4h-patient-example.patient.json#L1), which breaks the current SDK.

## Types of changes

## Checklist:

### DIFF
--- a/sdk-core/src/test/java/care/data4life/sdk/RecordServiceTest.kt
+++ b/sdk-core/src/test/java/care/data4life/sdk/RecordServiceTest.kt
@@ -63,11 +63,15 @@ import org.junit.After
 import org.junit.Assert
 import org.junit.Before
 import org.junit.Test
+import org.junit.internal.Classes.getClass
 import org.mockito.Mockito
 import org.mockito.Mockito.times
 import org.threeten.bp.LocalDate
 import org.threeten.bp.LocalDateTime
+import java.io.File
 import java.io.IOException
+import java.nio.charset.Charset
+import java.nio.charset.StandardCharsets
 
 class RecordServiceTest : RecordServiceTestBase() {
 
@@ -80,6 +84,10 @@ class RecordServiceTest : RecordServiceTestBase() {
     fun tearDown() {
         stop()
     }
+
+    private fun getResource(
+            resourceName: String
+    ): String = this::class.java.getResource(resourceName).readText(StandardCharsets.UTF_8)
 
     //region utility methods
     @Test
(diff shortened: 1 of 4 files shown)


sdk-core/src/test/java/care/data4life/sdk/RecordServiceTest.kt
```diff
@@ -823,6 +825,79 @@ class RecordServiceTest : RecordServiceTestBase() {
         inOrder.verifyNoMoreInteractions()
     }
 
+    @Test
+    @Throws(DataRestrictionException.UnsupportedFileType::class, DataRestrictionException.MaxDataSizeViolation::class)
+    fun `Given, checkDataRestrictions is called, with a Resource, which has non extractable Attachments, it returns without a failure`() {
+        // Given
+        val resourceStr = "{\n" +
+                "    \"resourceType\": \"Patient\",\n" +
```
wmontwe, line 833: I think the example payload should be shrinked to a minimal version, otherwise there is a lot of information that is not related to this check

wmontwe (changes requested):
Good fix 👍 Just the example payload is a little to big.

sdk-core/src/test/java/care/data4life/sdk/RecordServiceTest.kt
wmontwe: or the resource is moved to the resource folder

bitPogo: I pulled it out in the resource folder...However, since we will refactor the test soonish, I will invest a bit more time to think of a more generic example then, to decouple the test.

sdk-core/src/test/java/care/data4life/sdk/RecordServiceTest.kt
```diff
@@ -823,6 +833,22 @@ class RecordServiceTest : RecordServiceTestBase() {
         inOrder.verifyNoMoreInteractions()
     }
 
+    @Test
+    @Throws(DataRestrictionException.UnsupportedFileType::class, DataRestrictionException.MaxDataSizeViolation::class)
+    fun `Given, checkDataRestrictions is called, with a Resource, which has non extractable Attachments, it returns without a failure`() {
+        // Given
+        val resourceStr = getResource("/fhir4/s4h-patient-example.patient.json")
```
wmontwe, line 840: 👍

wmontwe (approved):
👍

PR #41, opened 2021-01-19
Fix versioning for releases and improve pull-request job

wmontwe:
## Description

Change the CI Jobs to properly generate versions based on the job executed

- Fix publish job by patching the versioning script to allow the branch name to be null, as we only need the tag
-  Change pull-request validation job to use the branch reference and remove custom checkout

## Types of changes

## Checklist:

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,5 +1,5 @@
 #
-# Copyright (c) 2020 D4L data4life gGmbH / All rights reserved.
+# Copyright (c) 2021 D4L data4life gGmbH / All rights reserved.
 #
 # D4L owns all legal rights, title and interest in and to the Software Development Kit ("SDK"),
 # including any intellectual property rights that subsist in the SDK.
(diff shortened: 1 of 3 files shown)


gradle.properties
```diff
@@ -14,26 +14,15 @@
...
-# org.gradle.parallel=true
-
-android.enableJetifier=true
-android.useAndroidX=true
-
-org.gradle.configureondemand=false
-
+# Gradle
 org.gradle.jvmargs=-Xmx1536m
+org.gradle.parallel=false
```
bitPogo, line 19: Why we don't run builds not any longer in parallel?

bitPogo (commented):
Just a question...but overall looks fine to me.

bitPogo (approved):

gradle.properties
wmontwe: it is flaky due to race conditions

PR #45, opened 2021-01-21
[NAT-496] Apply URI encoding on annotations and tags

bitPogo:
## Description
This PR aligns the SDK with its javascript sibling regarding the tag handling. However it remains currently misaligned regarding the actual behaviour in terms of the annotation handling.

## Motivation and Context
The SDK currently is misaligned with the javascript version regarding the annotation and tag handling, which will be fixed partially by this PR.

The TagHelper, which is now called TagEncryptionHelper, gets new auxiliary (`encode` and `decode`) which are used by the TagEncryptionService. Those methods do the actual formatting validation and apply the encoding/decoding. 
Alongside with the behaviour changes, the TagEncryptionService as well as the TagEncryptionHelper got now a more elaborate interface.

## How Has This Been Tested?
All behaviour changes are covered by (new) tests in regards to the unit and integration tests.
However the TagHelperTests had been relocated and renamed. The TagEncryptionTest are now using (mostly) mockk, since Mockito had some issues with the changes.

## Types of changes

## Checklist:


sdk-core/src/main/java/care/data4life/sdk/tag/TagEncryptionHelper.kt
```diff
@@ -0,0 +1,75 @@
...
+    }
+
+    @Throws(D4LException::class)
+    override fun encode(tag: String): String {
+        validateTag(tag)
+
+        return URLEncoder.encode(
+                tag.toLowerCase(Locale.US).trim(),
+                StandardCharsets.UTF_8.displayName()
+        ).map { char -> replaceSpecial(char) }.joinToString("")
```
alessioborraccino, line 69: Does this return a lowercased version of the percent encoded characters? When i was basic testing on the app,  the Web returned a lowercased version of the escaped characters ("%2d" instead of "%2D") and because of it i wasnt able to retrieve the key when searching with "d4l-donation-key" with the uppercased version.

bitPogo: There you go: https://github.com/d4l-data4life/hc-sdk-kmp/blob/2416e4658dec906a0ca9f1f30af6e425b80905e3/sdk-core/src/main/java/care/data4life/sdk/tag/TagEncryptionHelper.kt#L49

bitPogo: And the proof:
https://github.com/d4l-data4life/hc-sdk-kmp/blob/2416e4658dec906a0ca9f1f30af6e425b80905e3/sdk-core/src/test/java/care/data4life/sdk/tag/TagEncryptionHelperTest.kt#L141

PR #49, opened 2021-01-27
[NAT-496] Apply URI encoding on annotations and tags

bitPogo:
## Description
This PR aligns the SDK with its javascript sibling regarding the tag handling. However it remains currently misaligned regarding the actual behaviour in terms of the annotation handling.

## Motivation and Context
The SDK currently is misaligned with the javascript version regarding the annotation and tag handling, which will be fixed partially by this PR.

The TagHelper, which is now called TagEncryptionHelper, gets new auxiliary (`encode` and `decode`) which are used by the TagEncryptionService. Those methods do the actual formatting validation and apply the encoding/decoding. 
Alongside with the behaviour changes, the TagEncryptionService as well as the TagEncryptionHelper got now a more elaborate interface. Also TagEncryptionService got a refactoring and does not need any longer RxJava.

## How Has This Been Tested?
All behaviour changes are covered by (new) tests in regards to the unit and integration tests.
However the TagHelperTests had been relocated and renamed. The TagEncryptionTest are now using mockk, since Mockito had some issues with the changes.

## Types of changes

- [x] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:


sdk-core/src/main/java/care/data4life/sdk/tag/TagEncryptionService.kt
```diff
@@ -50,77 +66,76 @@ class TagEncryptionService @JvmOverloads constructor(
...
                 .map { encryptedTag -> decryptTag(tek, encryptedTag).blockingGet() }
+                .map { encodedTag -> tagHelper.decode(encodedTag) }
                 .filter { decryptedItem -> condition(decryptedItem) }
                 .toList()
                 .map { decryptedList -> transform(decryptedList) }
                 .blockingGet()
     }
 
+    @Throws(D4LException::class)
+    private fun decryptTag(key: GCKey, base64tag: String): Single<String> {
```
wmontwe, line 77: I would move the private members after the public accessors

sdk-core/src/main/java/care/data4life/sdk/tag/TagEncryptionService.kt
```diff
@@ -40,6 +43,19 @@ class TagEncryptionService @JvmOverloads constructor(
                 .blockingGet()
     }
 
+    @Throws(D4LException::class)
+    private fun encryptTag(key: GCKey, tag: String): Single<String> {
```
wmontwe, line 47: Even if you just moved the code, a shrinked version avoiding Single as suggestion. Error handling needs to be pushed up the stream or fail in place:

```
    @Throws(D4LException::class)
    private fun encryptTag(key: GCKey, tag: String): String {
        return listOf(tag.toByteArray())
                .map { plain -> cryptoService.symEncrypt(key, plain, IV) }
                .map { data -> base64.encodeToString(data) }
                .first()
    }
```

sdk-core/src/main/java/care/data4life/sdk/tag/TagEncryptionService.kt
```diff
@@ -19,16 +19,19 @@ import care.data4life.crypto.GCKey
 import care.data4life.crypto.error.CryptoException
 import care.data4life.sdk.CryptoService
 import care.data4life.sdk.lang.D4LException
+import care.data4life.sdk.tag.TaggingContract.Companion.TAG_DELIMITER
```
wmontwe, line 22: 👍

sdk-core/src/main/java/care/data4life/sdk/tag/TagEncryptionService.kt
```diff
@@ -50,77 +66,76 @@ class TagEncryptionService @JvmOverloads constructor(
...
 
     @Throws(IOException::class)
-    fun decryptTags(
+    override fun decryptTags(
             encryptedTags: List<String>
     ): HashMap<String, String> = decryptList(
             encryptedTags,
-            { d -> !d.startsWith(ANNOTATION_KEY) && d.contains(TaggingService.TAG_DELIMITER) },
-            { tagList: List<String> -> TagHelper.convertToTagMap(tagList) }
+            { decrypted -> !decrypted.startsWith(ANNOTATION_KEY) && decrypted.contains(TAG_DELIMITER) },
```
wmontwe, line 104: better readable 👍

sdk-core/src/test/java/care/data4life/sdk/RecordServiceCreationIntegration.kt
```diff
@@ -317,19 +315,23 @@ class RecordServiceCreationIntegration : RecordServiceIntegrationBase() {
...
         val tags = mapOf(
-                "partner" to "partner=TEST",
-                "client" to "client=TEST",
-                "fhirversion" to "fhirversion=3.0.1",
+                "partner" to "partner=$PARTNER_ID".toLowerCase(),
+                "client" to "client=${
+                    URLEncoder.encode(CLIENT_ID.toLowerCase(), StandardCharsets.UTF_8.displayName())
+                }",
+                "fhirversion" to "fhirversion=${
+                    "3.0.1".replace(".", "%2e")
```
wmontwe, line 323: This looks like a breaking change! If a FHIR version was `fhirversion=3.0.1` and is now `fhirversion=3%2e0%2e1`, we would have different tags after encryption which result in different results when querying the RecordService API. Is the JS also encoding the fhir version tag? For changing this we would need a migration mechanism, to keep existing datasets queryable.

sdk-core/src/test/java/care/data4life/sdk/RecordServiceIntegrationBase.kt
```diff
@@ -58,7 +58,10 @@ abstract class RecordServiceIntegrationBase {
 
     companion object {
         @JvmStatic
-        protected val CLIENT_ID = "TEST"
+        protected val PARTNER_ID = "B46"
+
+        @JvmStatic
+        protected val CLIENT_ID = "$PARTNER_ID#TEST"
```
wmontwe, line 64: 👍

sdk-core/src/test/java/care/data4life/sdk/tag/TagEncryptionServiceTest.kt
```diff
@@ -17,117 +17,122 @@ package care.data4life.sdk.tag
...
-                .symEncrypt(
-                        gcKey,
-                        "key=value".toByteArray(),
-                        IV
-                )
-        Mockito.doReturn(encryptedTag)
-                .`when`(mockBase64)
-                .encodeToString(symEncrypted)
-
+        every { tagHelper.encode(tag.second) } returns tag.second
```
wmontwe, line 63: 👍

wmontwe (changes requested):
Overall good 👍 

The impact of the change on the FHIR version number is a breaking change and needs to be validated. In case we go this route, we need to have a migration in place.

sdk-core/src/main/java/care/data4life/sdk/tag/TagEncryptionService.kt
bitPogo: Very good point...and I guess I actually what was the original intent now. The reverse function can be altered in the same way. Therefore we can totally get rid of RxJava in this class.
However the crypto service methods should be (in the future) then some sort async function.

sdk-core/src/main/java/care/data4life/sdk/tag/TagEncryptionService.kt
bitPogo: No problem...For me it's a bit easier to read it this way.

bitPogo (commented):
Good catches!
Regarding the breaking/migration: I am not entirely sure if we need in regards to tags since they should remain the same internally and the Decoder can handle also the "old" FHIR Version numbers. 
However what actually can break are annotations since the validation, lowercasing and the urlEncoding. (I will mark the patch accordingly) But annotations are not documented right now, so we did not actually propagate how to use them to a consumer. Unless they do not make something weird by now (like `+`or `%42`) we should be fine.

sdk-core/src/test/java/care/data4life/sdk/RecordServiceCreationIntegration.kt
bitPogo: JS applies the encoding in this way. See [here](https://github.com/d4l-data4life/js-sdk/blob/6e043843c525da93369db7c654c8515c8af6ac89/src/lib/stringUtils.ts#L2-L7), [here](https://github.com/d4l-data4life/js-sdk/blob/6e043843c525da93369db7c654c8515c8af6ac89/src/lib/taggingUtils.ts#L70-L76) and [here](https://github.com/d4l-data4life/js-sdk/blob/6e043843c525da93369db7c654c8515c8af6ac89/src/lib/taggingUtils.ts#L54-L56) .
My guess is that it supposed to overcome some compatibility issues regarding UTF-8/UTF-16 (but I am not sure about this). The queries itself should be fine since the tags and annotations are encoded and encrypted the very same way. (See [here](https://github.com/d4l-data4life/hc-sdk-kmp/blob/1b22b53622f569c70f97f70e4c32c9e5f4e8910e/sdk-core/src/main/java/care/data4life/sdk/RecordService.kt#L274-L280)).
However what I do not know (yet) is what the backend is doing, but since JS does it this way, I think we should be okish (but this actually my point of uncertainty).

wmontwe: The backend does nothing beside searching for exactly the representation of a tag the SDK is generating. 

New behaviour: `fhirversion=3.0.1` -> encode -> `fhirversion=3%2e0%2e1` -> encrypt -> `abzrAn37nmdnud8u=` -> base64 -> `Y2xpZW50PWI0NiUyM3Rlc3Q`

The old behaviour: `fhirversion=3.0.1` -> encrypt -> `cGFydG5lcj1iNDY=` -> base64 -> `ZmhpcnZlcnNpb249My=`

`Y2xpZW50PWI0NiUyM3Rlc3Q` is the identifier the backend is searching for and is not the same as `ZmhpcnZlcnNpb249My=`

The old behaviour is actually a bug as it should have been encoded too. But now we have Java/Android that generate `fhirversion=3.0.1` = `ZmhpcnZlcnNpb249My=` and iOS/JS where `fhirversion=3.0.1` = `Y2xpZW50PWI0NiUyM3Rlc3Q`

So we need to have a migration or compatibility in place to fix this bug.

sdk-core/src/main/java/care/data4life/sdk/tag/TagEncryptionService.kt
```diff
@@ -19,108 +19,110 @@ import care.data4life.crypto.GCKey
...
-        return Single
-                .fromCallable { tag.toByteArray() }
-                .map { d -> cryptoService.symEncrypt(key, d, IV) }
-                .map { data -> base64.encodeToString(data) }
-                .onErrorResumeNext {
-                    Single.error(
-                            CryptoException.EncryptionFailed("Failed to encrypt tag") as D4LException
-                    )
-                }
+    private fun encryptItem(key: GCKey, tag: String): String {
```
wmontwe, line 82: RxJava is gone 👍

wmontwe (changes requested):
Nice that some of RxJava is gone.

Could you please add to the changelog entry that it is a breaking change. Please also place a TODO in the migration section. To be filled when we have a migration strategy.

sdk-core/src/test/java/care/data4life/sdk/RecordServiceCreationIntegration.kt
bitPogo: As discussed: Agree!

wmontwe (approved):
👍

PR #50, opened 2021-02-04
Change EncryptedRecord, EncryptedKey and EncryptedKeyAdapter to Kotlin

bitPogo:
## Description
This simply converts the EncryptedRecord, EncryptedKeyAdapter  and EncryptedKey to Kotlin and adds an interface for the EncryptedRecord as well as for the EncryptedKey.

## Motivation and Context
The goal of this change is, besides going one step further towards Multiplattform, to make handling of tests more easier in the upcoming changes.

The Java classes have been converted into data classes and new interfaces had been added.
Since the EncryptedKey now has a factory some refactoring of the CryptoService was possible.

## How Has This Been Tested?
In order to verify the serialization still works, a new test set had been added for the EncryptedRecord.
Also since the EncryptedKey is capable of encoding/decoding its value some tests are added in this regard.

## Types of changes

## Checklist:


sdk-core/src/main/java/care/data4life/sdk/RecordService.kt
```diff
@@ -610,24 +612,24 @@ class RecordService(
 
         builder.setDataKey(dataKey)
 
-        if (record.encryptedAttachmentsKey != null) {
+        if (record.encryptedAttachmentsKey is EncryptedKey) {
```
wmontwe, line 615: if you use a value the compiler is able to check for not null and you don't need force !!

```
        val attachmentKey = record.encryptedAttachmentsKey
        if (attachmentKey is EncryptedKey) {
            builder.setAttachmentKey(
                    cryptoService.symDecryptSymmetricKey(
                            commonKey,
                            attachmentKey
                    ).blockingGet()
            )
        }
```

sdk-core/src/main/java/care/data4life/sdk/RecordService.kt
```diff
@@ -610,24 +612,24 @@ class RecordService(
...
+                            record.encryptedAttachmentsKey!!
                     ).blockingGet()
             )
         }
 
         return builder.build(
-                if (record.encryptedBody == null || record.encryptedBody.isEmpty()) {
-                    null// Fixme: This is a potential bug
-                } else {
+                if (record.encryptedBody is String && record.encryptedBody!!.isNotEmpty()) {
```
wmontwe, line 625: to avoid using force !!

```
        val body = record.encryptedBody
        return builder.build(
                if (body is String && body.isNotEmpty()) {
                    fhirService.decryptResource<T>(
                            dataKey,
                            tags,
                            body
                    )
                } else {
                    null// Fixme: This is a potential bug
                }
        ) as DecryptedBaseRecord<T>
```

sdk-core/src/main/java/care/data4life/sdk/network/model/EncryptedKey.kt
```diff
@@ -13,34 +13,6 @@
...
-        EncryptedKey that = (EncryptedKey) o;
-        return Objects.equals(encryptedKeyBase64, that.encryptedKeyBase64);
-    }
-
-    @Override
-    public int hashCode() {
-        return Objects.hash(encryptedKeyBase64);
-    }
-}
+data class EncryptedKey(val encryptedKey: String)
```
wmontwe, line 18: the power of Kotlin 🥇

wmontwe: maybe the name could reflect the encoding `encryptedKey` -> `base64Key` or `keyBase64`

don't know if it would be possible to enforce base64 encoding with a check here

sdk-core/src/main/java/care/data4life/sdk/network/model/EncryptedRecord.kt
```diff
@@ -0,0 +1,51 @@
...
+ * contact D4L by email to help@data4life.care.
+ */
+package care.data4life.sdk.network.model
+
+import care.data4life.sdk.network.model.NetworkModelContract.EncryptedRecord.Companion.DEFAULT_COMMON_KEY_ID
+import com.squareup.moshi.Json
+
+
+data class EncryptedRecord(
+        @field:Json(name = "common_key_id")
```
wmontwe, line 23: The Kotlin converter adds @field:Json and I asked myself why -> https://kotlinlang.org/docs/reference/annotations.html#annotation-use-site-targets

Seems legit

sdk-core/src/main/java/care/data4life/sdk/network/model/EncryptedRecord.kt
```diff
@@ -0,0 +1,51 @@
...
+        private var _commonKeyId: String?,
```
wmontwe, line 24: maybe instead of using `_commonKeyId` what about

```
        @Json(name = "common_key_id")
        override val commonKeyId: String = DEFAULT_COMMON_KEY_ID,
```
and remove 

```
    init {
        _commonKeyId = _commonKeyId ?: DEFAULT_COMMON_KEY_ID
    }

    override val commonKeyId: String
        get() = this._commonKeyId!!
```

wmontwe (approved):
Nice 👍 

Added some ideas for improvements

sdk-core/src/main/java/care/data4life/sdk/network/model/EncryptedKey.kt
bitPogo: A check is a bit difficult, since it is not really dividable from a normal string. But we can put a factory on top (since it is not really feasible to make it happen with Kotlins constructors in data classes) which does the actual encoding. That means we can pull this out of the CryptoService (which should not know something about encoding in my opinion). Also we might then also want to a `asDecoded` method/getter to this class (to pull the decoding also out of the CryptoService. What do you think?

sdk-core/src/main/java/care/data4life/sdk/network/model/EncryptedRecord.kt
bitPogo: This is indeed not pretty. But the original intend is a bit different. It seems it was assumed that on unmarshalling the Record it might happen that the commonKey is null. I am not sure if this can be the case. If not we even do not need a default value. However if it is the case, then the input needs to be nullable, I guess, and the default value will have not really an effect.

sdk-core/src/main/java/care/data4life/sdk/RecordService.kt
bitPogo: done

sdk-core/src/main/java/care/data4life/sdk/RecordService.kt
bitPogo: done

sdk-core/src/main/java/care/data4life/sdk/network/model/EncryptedKey.kt
wmontwe: 👍

sdk-core/src/main/java/care/data4life/sdk/network/model/EncryptedRecord.kt
wmontwe: Looking at the phdp-records service api the commonKeyId is mandatory, but when reviewing iOS it optional there too

wmontwe (approved):
👍

PR #51, opened 2021-02-11
Change RecordService#countRecords tests

bitPogo:
## Description
This refactors the test set for RecordService#countRecords. It also make some minor changes on the surface of the RecordService.

## Motivation and Context
Currently the tests of the SDK are not in a good shape. Also in the long run there is the need to replace mockito since it is troublesome in regards of non nullable parameters.
At last the performance of the tests is bad, which needs to be improved.

kotlin-test is now hooked up in the tests setup to enable more readable failure test.
Also some properties of the RecordService are now pointing to interfaces instead of objects and a typo has been fixed.
However the main focus is the test set of RecordService#countRecords which uses now mockk instead of mockito.

## How Has This Been Tested?
Since this is about tests, no further test sets are needed.

## Types of changes

## Checklist:

### DIFF
--- a/sdk-core/build.gradle.kts
+++ b/sdk-core/build.gradle.kts
@@ -54,6 +54,7 @@ dependencies {
         exclude(group = "care.data4life.hc-fhir-sdk-java", module = "hc-fhir-sdk-java")
     }
     testImplementation(Dependencies.Java.Test.junit)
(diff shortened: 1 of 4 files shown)


wmontwe (approved):
👍

PR #57, opened 2021-02-16
[SDK-523] Add countRecords support for Fhir4

bitPogo:
## Description
This adds the count functionality for Fhir4Records.
Additionally it fixes an issue encountered at countRecords while counting all Fhir3Records.

## Motivation and Context
Currently DataDonation is need of the count functionality for Fhir4.
This PR depends on #59.

This PR makes 4 major steps:
1. It adds a new function to the TaggingService(`tagVersion`) to enable version tagging via FhirVersion,
2. It refactors and fixes `countRecords` in the RecordService
3. It adds the Fhir4 functionality for count to the RecordService
4. It adds the Fhir4 functionality for count to the Fhir4Client.
5. It takes advantage of the behaviour which results from `DomainRefence` while tagging a query against the Backend.

## How Has This Been Tested?
All new methods have been covered by unittests. Additionally the exiting functional tests for count have been extended to also cover Fhir4.

## Types of changes

## Checklist:

### DIFF
--- a/sdk-core/src/main/java/care/data4life/sdk/tag/TaggingService.kt
+++ b/sdk-core/src/main/java/care/data4life/sdk/tag/TaggingService.kt
@@ -29,6 +29,7 @@ import care.data4life.sdk.tag.TaggingContract.Companion.TAG_UPDATED_BY_CLIENT
 import care.data4life.sdk.tag.TaggingContract.Companion.TAG_UPDATED_BY_PARTNER
 import care.data4life.sdk.wrapper.SdkFhirElementFactory
 import care.data4life.sdk.wrapper.WrapperContract
+import kotlin.collections.HashMap
 
 // TODO internal
 class TaggingService(
@@ -39,8 +40,8 @@ class TaggingService(
 
     private fun appendCommonDefaultTags(
             resourceType: String?,
-            oldTags: HashMap<String, String>?
-    ): HashMap<String, String> {
+            oldTags: Tags?
+    ): Tags {
         val tags = hashMapOf<String, String>()
         if (oldTags != null && oldTags.isNotEmpty()) {
             tags.putAll(oldTags)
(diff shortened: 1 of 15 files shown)


sdk-core/src/main/java/care/data4life/sdk/tag/TaggingService.kt
```diff
@@ -63,35 +63,48 @@ class TaggingService(
...
     ): HashMap<String, String> {
-        return hashMapOf<String, String>().also {
-            if (resourceType == null) {
-                it[TAG_APPDATA_KEY] = TAG_APPDATA_VALUE
-            } else {
-                it[TAG_RESOURCE_TYPE] = fhirElementFactory.getFhirTypeForClass(resourceType)!!
-                it[TAG_FHIR_VERSION] = fhirElementFactory.resolveFhirVersion(resourceType).version
+        val tags = hashMapOf<String, String>()
+
+        val version = if (resourceType == null) {
```
wmontwe, line 100: I think having a AppData type would be more robust than using null

sdk-core/src/main/java/care/data4life/sdk/wrapper/SdkFhirParser.kt
```diff
@@ -18,9 +18,9 @@ internal object SdkFhirParser : WrapperContract.FhirParser {
     }
 
     override fun toFhir4(resourceType: String, source: String): Fhir4Resource {
-        val clazz = fhirElement.getFhir4ClassForType(resourceType)!!
+        val clazz = fhirElement.getFhir4ClassForType(resourceType)
 
-        return fhir4Parser.toFhir(clazz, source)
+        return fhir4Parser.toFhir(clazz!!, source)
```
wmontwe, line 23: if `fhirElement.getFhir4ClassForType(resourceType)` could produce null, it makes sense to handle this case explicit. Using `!!` creates a NullPointerException but the Stacktrace for `!!` is not helpfull to debug later, no line number.

sdk-core/src/test/java/care/data4life/sdk/tag/TaggingServiceTest.kt
```diff
@@ -211,14 +211,50 @@ class TaggingServiceTest {
...
+
+        // When
+        taggingService.tagVersion(map, FhirContract.FhirVersion.FHIR_4)
+
+        assertTrue(map.containsKey(TAG_FHIR_VERSION))
+        assertEquals(FhirContract.FhirVersion.FHIR_4.version, map[TAG_FHIR_VERSION])
+    }
+
+    @Test
+    fun `Given, tagVersion is called with UNKNOWN and a map, it adds the appropriate TAG_FHIR_VERSION field`() {
```
wmontwe, line 247: should a unknown version be added as a tag?

wmontwe (dismissed):
👍 
 
and added some questions

sdk-core/src/main/java/care/data4life/sdk/wrapper/SdkFhirParser.kt
wmontwe: `!!` should only be used if Java Kotlin interop incorrectly reports a nullable type but it never could be null. If the value could be null, becuase of the programm logic and there needs to be a null handling instead.

bitPogo:
Please note I will rebase this patch on #59, since the mentioned PR makes it much easier to establish counting for Fhir3 and Fhir4 in the desired way.

sdk-core/src/main/java/care/data4life/sdk/tag/TaggingService.kt
bitPogo: Agree, but I will solve it a bit differently.

sdk-core/src/main/java/care/data4life/sdk/wrapper/SdkFhirParser.kt
bitPogo: This is just alignment with its twin method. `null` will only happen, if a consumer seeks to parse a DomainResource or other Resource we do not currently support.
However in principle I agree that we should tackle this in better way, but I see it in a separate task since this PR does not change the current behaviour in this particular case. Do you agree?

sdk-core/src/main/java/care/data4life/sdk/tag/TaggingService.kt
```diff
@@ -63,39 +63,48 @@ class TaggingService(
...
-                }
-            }
-            is Fhir4Resource -> appendCommonDefaultTags(resource.resourceType, oldTags).also {
-                if (!it.containsKey(TAG_FHIR_VERSION)) {
-                    it[TAG_FHIR_VERSION] = FhirContract.FhirVersion.FHIR_4.version
-                }
-            }
-            else -> appendCommonDefaultTags(null, oldTags).also {
-                it[TAG_APPDATA_KEY] = TAG_APPDATA_VALUE
-            }
```
bitPogo, line 81: I just like to mark, that the tagging check aka `!it.containsKey` is gone now.

sdk-core/src/main/java/care/data4life/sdk/wrapper/SdkFhirParser.kt
wmontwe: I commented, because the issue was just moved from on property to the other

wmontwe:
> Please note I will rebase this patch on #59, since the mentioned PR makes it much easier to establish counting for Fhir3 and Fhir4 in the desired way.

wmontwe (approved):
👍

PR #58, opened 2021-02-17
[SDK-525] Add compatibility layer for legacy Tags and Annotation

bitPogo:
## Description
This adds new layer to the SDK to support legacy resources tagged without encoding.

## Motivation and Context
A legacy problem where the android/kmp SDKs was not percent encoding tags/annotations (differently from iOS and Js) created incoherent tagging for resources.
This resulted in production with resources tagged with "fhirversion=3.0.1" and "fhirversion=3%2e0%2e1", resulting in different tags for the same version, and since the introduction of fhir 4, using such tags to distinguish which fhir versions we want to retrieve will fail to produce a coherent result (since different platform are tagging resources differently).
This depends on #57 .

The SDK got now a new Layer which is responsible for compatibility issues.
The TagEncryptionHelper and TaggingService got some auxiliary methods.
The RecordService has now 2 constructors (until we introduce a factory) to make testing easier and mark the intend explicitly. Also the RecordService now calls on `searchRecords` and `_countRecords` the new CompatibilityService.
The calls fulfilling following properties:
- On Fhir3 & Fhir4 it calls always twice on `searchRecords` and `_countRecords` (once with the new form, once with the legacy form)
- On arbitrary data it calls once if tags and annotations in both forms equal each other otherwise it calls twice the api.
- `_countRecords` is not supported (yet) for arbitrary data.

## How Has This Been Tested?
All changes are covered by unit tests and the integration test had been adjusted accordingly to cover the behaviour.

## Types of changes

## Checklist:


sdk-core/src/main/java/care/data4life/sdk/ApiService.java
```diff
@@ -241,7 +241,8 @@ Completable uploadTagEncryptionKey(String alias, String userId, String encrypted
         return service.createRecord(alias, userId, (EncryptedRecord) encryptedRecord);
     }
 
-    Observable<List<EncryptedRecord>> fetchRecords(String alias,
+    // TODO remove public
+    public Observable<List<EncryptedRecord>> fetchRecords(String alias,
```
alessioborraccino, line 245: What's this about? Does public means it will be visible on the client apis like in swift?

sdk-core/src/main/java/care/data4life/sdk/RecordService.kt
```diff
@@ -262,20 +292,20 @@ class RecordService(
     }
 
     private fun encryptTagsAndAnnotations(
-            painTags: HashMap<String, String>,
+            plainTags: HashMap<String, String>,
             plainAnnotations: List<String>
```
alessioborraccino, line 296: what does plain mean? not encoded?

sdk-core/src/main/java/care/data4life/sdk/fhir/Fhir4RecordClient.kt
```diff
@@ -75,4 +87,21 @@ internal class Fhir4RecordClient(
                 }
         return handler.executeSingle(operation, callback)
     }
+
+    override fun <T : Fhir4Resource> count(
```
alessioborraccino, line 91: on iOS we have a usable version of count with the same signature that doesnt pass the resourceType (uses a default argument of Fhir4Resource). So its not nullable but its omittable either way.  is it possible to create the same here?

sdk-core/src/main/java/care/data4life/sdk/migration/CompatibilityService.kt
```diff
@@ -0,0 +1,124 @@
...
+import care.data4life.sdk.ApiService
+import care.data4life.sdk.network.model.NetworkModelContract.EncryptedRecord
+import care.data4life.sdk.tag.Annotations
+import care.data4life.sdk.tag.TaggingContract
+import care.data4life.sdk.tag.Tags
+import io.reactivex.Observable
+import io.reactivex.Single
+import io.reactivex.functions.BiFunction
+
+class CompatibilityService internal constructor(
```
alessioborraccino, line 28: I think this name is a bit too generic. Compatibility of what? is it going to be used for as a catch all for all legacy compatibility issues? Maybe would be nice if it was more specific, maybe adding the word Record somewhere

sdk-core/src/main/java/care/data4life/sdk/migration/CompatibilityService.kt
```diff
@@ -0,0 +1,124 @@
...
+            tagEncryptionService.encryptTags(plainTags)
+                .also { encryptedTags ->
+                    encryptedTags.addAll(
+                        tagEncryptionService.encryptAnnotations(plainAnnotations)
+                    )
+                }
+        )
+    }
+
+    private fun callOnlyOnce(
```
alessioborraccino, line 52: in swift, when we have boolean func/props we write those as 3rd person singular, eg. `needsToBeCalledOnce` instead of `callOnlyOnce`, mostly because being a function, it might look like you are actually calling something only once and then maybe returning some boolean result.

In this actual case, being tags the parameters of the function, i would make it clear, as in
`fun tagsNeedDoubleCall(plainTags: tags, encryptedTags: Tags): Bool {} `
signaling that tags are the decision, but i leave it to you as i am not as familiar with kotlin styles

sdk-core/src/main/java/care/data4life/sdk/RecordService.kt
bitPogo: Plain mean in this case not encoded and not encrypted...but you point me here to something else thx!

sdk-core/src/main/java/care/data4life/sdk/ApiService.java
bitPogo: Since the new Service lives not in the same package but needs access I had to change the visibility. `public` means it is [visible](https://docs.oracle.com/javase/tutorial/java/javaOO/variables.html) to all other classes.

sdk-core/src/main/java/care/data4life/sdk/fhir/Fhir4RecordClient.kt
bitPogo: Maybe lets split the concerns, and lets see iff I got them correctly:
1. is `resourceType` nullable? - no it is not (since in Kotlin you have to declare nullable values explicitly). Hence consumer has to delegate a certain value. 
2. make `resourceType` omittable. Yeah we could do that. But the question is - what is a reasonable default value? `DomainResource` maybe, but this would lead to a countAll behaviour, which we discussed in a other place.

sdk-core/src/main/java/care/data4life/sdk/migration/CompatibilityService.kt
bitPogo: Agreed

sdk-core/src/main/java/care/data4life/sdk/migration/CompatibilityService.kt
bitPogo: Maybe `needsDoubleCall`?

alessioborraccino: Sure

sdk-core/src/main/java/care/data4life/sdk/fhir/Fhir4RecordClient.kt
alessioborraccino: I mean, from API parity prospective, on the public interface of iOS we have 3 count methods (one for each case, fhir3/4 and appdata)  and the fhir ones look like:

```
public func countFhirStu3Records<R: FhirStu3Resource>(of type: R.Type = R.self,
                                                          annotations: [String] = [],
                                                          queue: DispatchQueue = responseQueue,
                                                          completion: @escaping ResultBlock<Int>) {
        fhirService
            .countFhirRecords(of: type, annotations: annotations)
            .complete(queue: queue, completion)
    }
```

so if someone wants to count all fhirStu3 records can just write:
`countFhirStu3Records { count in ... }`  and it automatically uses R.self as type (which is DomainResource, which recordService turns into a no-parameter, effectively doing a countAllFhir3Records)

I would like the android api to have the same structure if possible

bitPogo: Ah...I think I got it new... `{...}` calls the completion, right?
Technically we can do it in the future, but for now we are bounded to the legacy client. I agree that our API should be aligned as possible crossplattform. However I do not see it in this particular task. Also the team did not decided how to handle the parallelisms in the future. For example to make something similar in Kotlin we need to change the SDKContract that the callback is a closure and not a `Callback`.

sdk-core/src/main/java/care/data4life/sdk/migration/CompatibilityService.kt
bitPogo: done

sdk-core/src/main/java/care/data4life/sdk/migration/CompatibilityService.kt
bitPogo: done

sdk-core/src/main/java/care/data4life/sdk/fhir/Fhir4RecordClient.kt
bitPogo: @wmontwe maybe you have opinion as well? Its about implementing the count all behaviour via default parameter. What you like prefer more having a dedicated function `countAllSomething` or just overload `countSomething`.
The overload could or an alias `countAllSomething` could be added here.
I guess that has also some implication for a search all for `Something`.

* replace `Something` with a concrete FHIR type or with Data

wmontwe: Hmm, I think self reference of a Generic type is not possible with Kotlin. Maybe I'm wrong. Also we need to ensure Java interop, which would require moving the parameter at last position to be able to create an overloaded function.

But I like both approaches:
* `countAll` is more intuitive, but broader API
* `count` with defaulting to DomainResource is slick, but maybe not as easy to grasp

`countAll` still leaves us with a client using `count` with `DomainResource`

wmontwe:
At the moment this is impossible to review as the base of this branch was rebased and the changeset includes both.

sdk-core/src/main/java/care/data4life/sdk/RecordService.kt
```diff
@@ -287,27 +318,20 @@ class RecordService(
...
                 }
                 .flatMapIterable { it }
                 .map { encryptedRecord -> decryptRecord<T>(encryptedRecord, userId) }
-                .let {
-                    if (resourceType == null) {
-                        it
-                    } else {
-                        it.filter { decryptedRecord -> resourceType.isAssignableFrom(decryptedRecord.resource::class.java) }
-                    }.filter { decryptedRecord -> decryptedRecord.annotations.containsAll(annotations) }
-                }
```
bitPogo, line 310: Just to make this change obvious.

bitPogo:
> At the moment this is impossible to review as the base of this branch was rebased and the changeset includes both.

Sry - during the attempt of merging #59 into #57 (and then into here) many stuff had been overwritten which made it really hard to figure out why some test broke, since the changes worked closely to each other. So rebase was much easier then merge, which lead to this.

sdk-core/src/main/java/care/data4life/sdk/fhir/Fhir4RecordClient.kt
bitPogo: @ouafae31

sdk-core/src/main/java/care/data4life/sdk/migration/RecordCompatibilityService.kt
```diff
@@ -0,0 +1,125 @@
...
+                                    tagEncryptionService.encryptAnnotations(plainAnnotations)
+                            )
+                        }
+        )
+    }
+
+    private fun needsDoubleCall(
+            encodedAndEncryptedTags: List<String>,
+            encryptedTags: List<String>
+    ): Boolean = encodedAndEncryptedTags.sorted() != encryptedTags.sorted()
```
wmontwe, line 56: Could you explain how that works? I don't get it.

wmontwe: Also the migration logic bleeds into other parts of SDK -> TagEncryptionService. Ideally it is located in one place that could be easily removed when not needed anymore. The need for `@Deprecated` is also a strong sign for missing decoupling.

wmontwe (changes requested):
The migration logic bleeds into the code base. It should be decoupled.

sdk-core/src/main/java/care/data4life/sdk/tag/TagEncryptionService.kt
```diff
@@ -32,23 +31,30 @@ class TagEncryptionService @JvmOverloads constructor(
...
-    override fun encryptTags(tags: HashMap<String, String>): MutableList<String> {
+    override fun encryptAndEncodeTags(tags: Tags): MutableList<String> {
         return tags
                 .map { entry -> entry.key + DELIMITER + tagHelper.encode(entry.value) }
-                .let { encryptList(it) }
+                .let { encodedTags -> encryptList(encodedTags) }
+    }
+
+    @Throws(IOException::class)
+    override fun encryptTags(tags: Tags): MutableList<String> {
```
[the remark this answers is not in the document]
bitPogo: This method should go away as soon as we managed the migration. In other words I like to make obvious (on a higher level) that this method should be there.

sdk-core/src/main/java/care/data4life/sdk/tag/TaggingContract.kt
```diff
@@ -16,35 +16,41 @@
...
         fun getTagsFromType(resourceType: Class<out Any>): Tags
-
     }
 
     interface EncryptionService {
         @Throws(IOException::class)
+        fun encryptAndEncodeTags(tags: Tags): MutableList<String>
+
+        @Throws(IOException::class)
+        @Deprecated("This method should only be used for migration purpose.")
```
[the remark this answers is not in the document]
bitPogo: You mean something linke `migration`?

sdk-core/src/test/java/care/data4life/sdk/RecordServiceCountRecordsTest.kt
```diff
@@ -47,9 +46,10 @@ class RecordServiceCountRecordsTest {
     private lateinit var attachmentService: AttachmentContract.Service
     private lateinit var errorHandler: SdkContract.ErrorHandler
     private lateinit var tags: HashMap<String, String>
-    private lateinit var encryptedTags: MutableList<String>
     private val defaultAnnotations = listOf<String>()
-    private lateinit var encryptedAnnotations: MutableList<String>
+
+    // mark
```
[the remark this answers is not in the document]
bitPogo: Good catch...thats a leftover (just a jump mark)

sdk-core/src/main/java/care/data4life/sdk/migration/RecordCompatibilityService.kt
bitPogo: > Could you explain how that works? I don't get it.
Sure!
The context can be found in [here](https://gesundheitscloud.atlassian.net/secure/RapidBoard.jspa?rapidView=109&projectKey=SDK&modal=detail&selectedIssue=SDK-534).

1.  the compatibility layer method gets a call
2. the method creats 2 version of the tags (one with url encoding one without)
3. then it checks if the 2 versions are equal (´needsDoubleCall`) since in case of arbitrary data we can encounter the case that both are equal (for example there no annotations or the annotations are [a-z]+)
We need this since otherwise we get the very same record twice.
The `sort` is arguably a little overdo in its given context now, but in my opinion is more 'safe', since context can change quickly.
4. the actual api call is executed
5. the results are concatenated

bitPogo: > Also the migration logic bleeds into other parts of SDK -> TagEncryptionService, RecordService, ... Ideally it is located in one place that could be easily removed when not needed anymore. The need for @deprecated is also a strong sign for missing decoupling.

I see your point and I am open to suggestions.
In terms of the TagEncryptionService: The root cause for this is the coupling of the encoding and encrypting, which could be done separately. However the drawback is we loop twice over the tags then and in the compatibility context 3 times. My intension was to avoid that. What I also want to avoid is some sort of feature flag like iOS does it, since feature flags stay much longer than deprecated methods from my experience.
Also an alternative is to make `encryptList` public/internal, which is also not pretty from my perspective, and pull `encryptTags` out and place it here.

In terms of the RecordService: I found a bit odd to have encrypt and append for both in 2 places, but this is only a matter of taste. But what we can do is to introduce this encrypt&append to the TagEncryptionService and get rid of it entirely in RecordService. But this also depends on the way we deal with TagEncryptionService.

So what do you think?

bitPogo:
As requested - I introduces a new Annotation and moved the legacy encryption into the new layer. Also I as discussed the RecordService calls now a bundle encryption method for tags and annotations.

wmontwe (dismissed):
👍

wmontwe (approved):
👍

PR #59, opened 2021-02-17
[NAT-548] Fix NullPointerException for DomainResource

bitPogo:
…omType

## Description
This fixes the current occurrence of a Nullpointer Exception while calling `fetchRecords` with a `DomainResource`.

## Motivation and Context
When calling `fetch/search` and using a `DomainResource` the SDK fails with a NullPointerException, which relates to the transition from Java to Kotlin.
Also allowed the former behaviour to use the translation from `DomainResource` to `null` in SDK to query all related Records.

This simply changes the mandatory String in the TaggingService#getTagFromType to optional and only tags the type if a resourceType is present.

## How Has This Been Tested?
2 new unit tests had been added for the TaggingService for Fhir3 and Fhir4. 
Also 2 new integration tests are in place to proof DomainResource is now working on `fetchRecords` for Fhir3 and Fhir4.

## Types of changes

## Checklist:

### DIFF
--- a/CHANGELOG.adoc
+++ b/CHANGELOG.adoc
@@ -36,6 +36,7 @@ toc::[]
 
 * RecordService#deleteRecord invocation had mixed user and resource id.
 * RecordService#fetchRecords invocation had mixed user and resource id.
+* NullPointerException when using DomainResource as resourceType for fetch/search.
 
 === Bumped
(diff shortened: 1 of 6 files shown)


wmontwe (approved):
👍

PR #61, opened 2021-02-19
Add DateTimeFormatter Wrapper

bitPogo:
## Description
This encapsulates the current date-time formatter logic in a dedicated place.

## Motivation and Context
This is about 2 things:
1. there are at least 2 places which are dealing with date-time formatting, which they are not supposed as their primary task.
2. since the project seeks to be getting more and more multiplattform, this is a small step into this direction.

This PR depends on #58.

## How is it being implemented?
This adds a new class - the `SdkDateTimeFormatter` and its interface (in `WrapperContract`).
Also it introduces an interface for `Meta` and forms a new `ModelContract` to get rid of the `definitions` package in `model`.
Additionally it removes all DateFormatting related stuff from the `RecordService` as well as from the `RecordMapper`.

## How Has This Been Tested?
All existing unit test had been adjusted to the current structure. 
A new test set had been introduced for the `DateTimeFormatter`.
Also new module tests had been added to also cover `startDate` and `endDate` for searchRecords.  

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [ ] I have updated the changelog accordingly.
- [x] I have added tests to cover my changes.
- [x] All new and existing tests passed.

PR #63, opened 2021-03-16
Refactor RecordServiceAdditionalResourceTypeTest

bitPogo:
## Description

This refactors and renames RecordServiceAdditionalResourceTypeTest.

## Motivation and Context

This depends on #61.
The currently RecordServiceAdditionalResourceTypeTest is a very bad shape - it is slow, hard to understand and applied only on Fhir3.

This makes a complete refactoring of the test set and labels it as a module test.
The tests are now run as a parameterized test, so it is more easier to add new resource types and let them run through all tests. Also the resources are not builded on runtime - they now loaded from the `resources` folder.
The attachment pdfs are also build only once, since they caused a massiv memory problem.
The tests are now running on Fhir4 resources as well.
Also it decouples the tests from the rest of the RecordServiceTest, which decreases the runtime of the tests by a lot.

## Types of changes

## Checklist:


ouafae31 (dismissed):
RecordServiceAdditionalResourceTypeTest LGTM.

sdk-core/src/test/java/care/data4life/sdk/RecordServiceAdditionalResourceTypeModule.kt
```diff
@@ -942,8 +950,8 @@ class RecordServiceAdditionalResourceTypeModule {
                             SdkAttachmentFactory.wrap(firstAttachment),
                             SdkAttachmentFactory.wrap(secondAttachment)
                     ),
-                    any(),
-                    any()
+                    attachmentKey,
+                    USER_ID
             )
```
ouafae31, line 955: Why is it important to remove any() ?

ouafae31 (approved):
I would prefer RecordServiceAdditionalResourceTypeModule to have test suffix to find it easily.
Good refactoring that allowed us to test all cases for all resourceType but unfortunately the readability of the each test with condition/consequence was reduced.

sdk-core/src/test/java/care/data4life/sdk/RecordServiceAdditionalResourceTypeModule.kt
bitPogo: `any()` in my opinion is never a good choice since you accept all payloads, since in a test environment the parameter should be known (except in some edge cases).  If you define the parameter specifically, the tests will not run and fail, which leads to much less problems with the test themselves.
Here I simply overlooked it the first time.

bitPogo:
> I would prefer RecordServiceAdditionalResourceTypeModule to have test suffix to find it easily.
Good refactoring that allowed us to test all cases for all resourceType but unfortunately the readability of the each test with condition/consequence was reduced.

Can you elaborate a bit more, what you mean by that?

ouafae31:
I just meant RecordServiceAdditionalResourceTypeModuleTest is better because I always expect '"test" suffix to differentiate between the class and its test.

ouafae31 (approved):
LGTM

PR #71, opened 2021-03-29
Pull RecordServiceAttachmentUpdateTest out of RecordServiceUploadsUpdatesDownloadsTest

bitPogo:
## Description

This is the 2nd of 3 PRS, which breaks the RecordServiceUploadsUpdatesDownloadsTest down.

## Motivation and Context

Currently RecordServiceUploadsUpdatesDownloadsTest is not in a proper state.
This is a successor of #66 and depends on #70.

## How is it being implemented?

This breaks out the update part out of the RecordServiceUploadsUpdatesDownloadsTest.
However a minor change had been made to RecordService#updateData to adjust it with the upcoming PR, which handles the upload of Attachments. It now will fetch the AttachmentKey if the Record has currently no attachment key attached.

## How Has This Been Tested?

A new test had been added to cover the new behaviour. However this is a major rewrite anyway.

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [ ] I have updated the changelog accordingly.
- [x] I have added tests to cover my changes.
- [x] All new and existing tests passed.

PR #66, opened 2021-03-22
Refactor RecordService tests in terms of attachment upload/download

bitPogo:
## Description

This refactors the RecordServiceUploadsUpdatesDownloadsTest set.
Please note, while doing the refactoring a bug was encountered and fixed.

## Motivation and Context

Currently RecordServiceUploadsUpdatesDownloadsTest is not in a proper state.

## How is it being implemented?

This breaks down RecordServiceUploadsUpdatesDownloadsTest in 3 separated sets, to tackle the size of the original.
It now makes proper testing in terms of Fhir3 and Fhir4.
Additionally it removes the corresponding legacy methods from the RecordService and makes some minor changes to RecordService to improve the readability of `updateData` and `uploadData`.
Since a bug was encountered in `downloadData`, this changes the method accordingly.
Also the module tests had been removed and will be reintroduced once #63 is merged.

## How Has This Been Tested?

A new unittest had been added to cover the bug. Besides that it makes a rewrite of the existing tests.

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have updated the changelog accordingly.
- [x] I have added tests to cover my changes.
- [x] All new and existing tests passed.

PR #72, opened 2021-03-29
Pull RecordServiceAttachmentUploadTest out of RecordServiceUploadsUpdatesDownloadsTest

bitPogo:
## Description

This is the 3rd of 3 PRS, which breaks the RecordServiceUploadsUpdatesDownloadsTest down.
Currently RecordServiceUploadsUpdatesDownloadsTest is not in a proper state.
This is a successor of #66 and depends on #71.

## Motivation and Context

Currently RecordServiceUploadsUpdatesDownloadsTest is not in a proper state.

This breaks out the update part out of the RecordServiceUploadsUpdatesDownloadsTest.
It removes internal legacy methods, which are not longer need and removes the corresponding tests as well.

## How Has This Been Tested?

This is a rewrite of tests which check the upload attachment logic.

## Types of changes

## Checklist:


sdk-core/src/main/java/care/data4life/sdk/RecordService.kt
```diff
@@ -124,11 +124,6 @@ class RecordService internal constructor(
             )
     )
 
-    @Deprecated("Deprecated with version v1.9.0 and will be removed in version v2.0.0")
```
wmontwe, line 127: 👍

sdk-core/src/main/java/care/data4life/sdk/RecordService.kt
```diff
@@ -913,13 +908,9 @@ class RecordService internal constructor(
         val resource = record.resource
 
         if (!fhirAttachmentHelper.hasAttachment(resource)) return record
-        val attachments = fhirAttachmentHelper.getAttachment(resource) as List<Any?>?
+        val attachments = fhirAttachmentHelper.getAttachment(resource)
                 ?: return record
 
-        if (record.attachmentsKey == null) {
```
wmontwe, line 919: Why is this removed? Isn't there a case the ther is no attachment key yet?

bitPogo: It is actually not removed, it is just moved (as mentioned in the PR before) into `resolveAttachmentKey`, which is called by `uploadAttachmentsOnDemand `.

wmontwe (approved):
👍

sdk-core/src/main/java/care/data4life/sdk/RecordService.kt
wmontwe: 👍

PR #76, opened 2021-04-07
Wrap up RecordServiceTest refactoring

bitPogo:
## Description

This wraps up the refactoring of the unit tests sets of the
`RecordService`.

## Motivation and Context

Since all test sets are now decoupled, the setup phase can be approached differently to gain more speed.
Also The former `RecordServiceTestBase` can now act as a provider.

Basically all `RecordService` unit test sets do not init all basic mocks (like the apiService) always fresh, instead they now do it once and clear them before each test to gain more speed.
The `RecordServiceTestBase` acts now as a data provider. Also outdated code (like mockito) was removed.
It also removes legacy code from the `RecordService` itself.

## Types of changes

## Checklist:


sdk-core/src/main/java/care/data4life/sdk/RecordService.kt
```diff
@@ -124,11 +124,6 @@ class RecordService internal constructor(
             )
     )
 
-    @Deprecated("Deprecated with version v1.9.0 and will be removed in version v2.0.0")
```
wmontwe, line 127: 👋

wmontwe (approved):
👍

PR #83, opened 2021-04-08
CryptoSecureStore.java to Kotlin

bitPogo:
## Description

This converts the CryptoSecureStore into Koltin.

## Motivation and Context

The project moves more and more to KMP, therefore everything needs to be Kotlin.
See [SDK-562](https://gesundheitscloud.atlassian.net/browse/SDK-562) and [SDK-564](https://gesundheitscloud.atlassian.net/browse/SDK-564).

This converts the CryptoSecureStore into Koltin.
In addition it adds a new contract and 2 new interfaces alongside (for the CryptoSecureStore and the KeyFactory.
It makes also minor changes to the CryptoService to reflect the changes.

## How Has This Been Tested?

No new tests had been added and existing tests had not been altered.

## Types of changes

## Checklist:


sdk-core/src/main/java/care/data4life/sdk/CryptoSecureStore.kt
```diff
@@ -0,0 +1,102 @@
...
+            null,
+            key.keyVersion
+        )
+        val json = moshi.adapter(ExchangeKey::class.java).toJson(exchangeKey)
+        secureStore.addData(alias, json.toCharArray())
+    }
+
+    @Throws(IOException::class)
+    override fun getExchangeKey(alias: String): ExchangeKey {
+        val data = secureStore.getData(alias) ?: CharArray(0)
```
wmontwe, line 96: Why is this returning an empty array in case the store doesn't contain data?

sdk-core/src/main/java/care/data4life/sdk/CryptoSecureStore.kt
```diff
@@ -0,0 +1,102 @@
...
+    private val secureStore: SecureStoreContract.SecureStore
+) : CryptoContract.CryptoSecureStore {
+    private val moshi: Moshi = if (moshi is Moshi) moshi else Moshi.Builder().build()
+
+    override fun clear(): Unit = secureStore.clear()
+
+    override fun storeSecret(
+        alias: String,
+        secret: CharArray
+    ): Unit = secureStore.addData(alias, secret)
```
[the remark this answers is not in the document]
bitPogo: Otherwise the Compiler does not expect the `=`.

sdk-core/src/main/java/care/data4life/sdk/CryptoSecureStore.kt
```diff
@@ -0,0 +1,102 @@
...
+import care.data4life.sdk.crypto.CryptoContract
+import care.data4life.sdk.lang.D4LException
+import care.data4life.securestore.SecureStoreContract
+import com.squareup.moshi.Moshi
+import com.squareup.moshi.Types
+import java.io.IOException
+import java.lang.reflect.Type
+
+class CryptoSecureStore @JvmOverloads constructor(
+    moshi: Moshi? = null,
```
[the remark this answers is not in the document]
bitPogo: I guess this was introduces in the original due to testing purposes (see [here](https://github.com/d4l-data4life/hc-sdk-kmp/blob/main/sdk-core/src/test/java/care/data4life/sdk/CryptoSecureStoreTest.java#L72)).

sdk-core/src/main/java/care/data4life/sdk/CryptoSecureStore.kt
bitPogo: Good catch! The original would simply passthrough the return value, which can be null.`fromJson` does not allow a null as parameter. But if we delegate an empty CharArray the old behaviour is preserved (sort of).
However nicer would be to fail in here then in moshi, but this is a change when can postpone until we touch the tests.

sdk-core/src/main/java/care/data4life/sdk/CryptoSecureStore.kt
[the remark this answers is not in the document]
bitPogo: Now I got it - I moved it up into the constructor.

sdk-core/src/main/java/care/data4life/sdk/CryptoSecureStore.kt
wmontwe: hmm, null was a testable error case. What about?
```
    @Throws(IOException::class)
    fun getExchangeKey(alias: String): ExchangeKey {
        val json = secureStore.getData(alias)
            ?: throw (DecryptionFailed("Failed to decrypt data"))

        return moshi.adapter(ExchangeKey::class.java).fromJson(json.toString())
            ?: throw (DecryptionFailed("Failed to decrypt data"))
    }
```

bitPogo: Done

wmontwe (approved):
👍

PR #86, opened 2021-04-09
Rewrite RecordServiceCreateIntegrationTest

bitPogo:
## Description

This PR is the first of a small series to improve the current module tests.
This set makes an rewrite of the RecordServiceCreationIntegration, which includes:
* rename it (the more appropriate terminology is module test then integration test)
* it aligns the setup phase to the new established pattern
* it decouples the test set from the other module tests
* it now makes use of the `GenericDataTestProvider`
* it pulls out most shared behaviour of the `CryptoService` to a helper
* it uses now templates for resources instead of setting them up on Runtime
* it utilises already existing sample attachments for tests instead of mocks.
* it reduces the boilerplate in the test set itself by a lot
* it solves certain issues of the capability of the tests themselves (for example it was not possible to divide between the encryption and decryption phase in terms of the `EncryptionKeys`)

Additionally this makes more elaborate tests possible, which is already used by this changes.

However there are major tradeoffs, which must to be mentioned:
It makes the flow of the test harder to understand, while increasing its readability. This is due to the fact, that this pulls out shared behaviour, while also separating the encryption and decryption. Eventually this leads to big function signatures, which are rather pretty, but necessary.
But this might be getting better when extending the tests to include the crypto as well.

## Motivation and Context

Currently the module tests (former known as integration test) are hard to maintain, since behaviour changes must done several times and the boilerplate for each test set is huge. This also leads to limited capabilities of the tests and a relatively static nature. 

As already mentioned it decouples the test set from the other module tests by introducing a new module test helper, which now holds shared faked behaviour and utilising the existing `GenericDataProvider`.

## Types of changes

## Checklist:


bitPogo:
Abandoned in favour of #89.

PR #90, opened 2021-04-11
Rewrite RecordService#updateRecord module test

bitPogo:
## Description

This patch makes an rewrite of the RecordServiceCreationIntegration, which includes:
* rename it (the more appropriate terminology is module test then integration test)
* it aligns the setup phase to the new established pattern
* it decouples the test set from the other module tests
* it now makes use of the DataTestProvider
* it uses now templates for resources instead of setting them up on Runtime
* it utilises already existing sample attachments for tests instead of mocks.
* it reduces the boilerplate in the test set itself 
* it solves certain issues of the capability of the tests themselves (for example it was not possible to divide between the  encryption and decryption phase in terms of the EncryptionKeys)
* it extends the already existing tests

This changes depending on #89.

## Motivation and Context

Currently the module tests (former known as integration test) are hard to maintain, since behaviour changes must done several times and the boilerplate for each test set is huge. This also leads to limited capabilities of the tests and a relatively static nature.
See [SDK-560](https://gesundheitscloud.atlassian.net/browse/SDK-560).

As already mentioned it decouples the test set from the other module tests by introducing a new module test helper and using the introduced `CryptoServiceFake`.

## Types of changes

## Checklist:


wmontwe (approved):
👍

PR #98, opened 2021-04-19
[SDK-564] Add UserService interface

bitPogo:
## Description

This simply adds an interface for the existing `UserService`.
See also [SDK-564](https://gesundheitscloud.atlassian.net/browse/SDK-564)

## Motivation and Context

In order to decouple implementation, this change is made.

## How is it being implemented?

Simply it adds an interface.

## How Has This Been Tested?

No new tests are needed.

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [ ] I have updated the changelog accordingly.
- [ ] I have added tests to cover my changes.
- [x] All new and existing tests passed.

PR #99, opened 2021-04-19
[SDK-548] Add UserId to Client

bitPogo:
## Description

This exposes the UserId in the clients.

## Motivation and Context

This is part of [SDK-548](https://gesundheitscloud.atlassian.net/browse/SDK-548) and depends on #98.

It adds a new private constructor to the `BaseClient`, which gets an additional parameter `userId`, which is fetch from the `UserService` in the public constructor.

## How Has This Been Tested?

A new test set for the BaseClient had been added, as well as tests on the concrete clients, to satisfy the requirements of the mentioned ticket.

## Types of changes

## Checklist:


sdk-core/src/main/java/care/data4life/sdk/BaseClient.kt
```diff
@@ -16,54 +16,66 @@
...
 import care.data4life.sdk.auth.AuthClient
+import care.data4life.sdk.auth.AuthContract
 import care.data4life.sdk.auth.UserService
 import care.data4life.sdk.call.CallHandler
 import care.data4life.sdk.data.DataRecordClient
 import care.data4life.sdk.fhir.Fhir4RecordClient
 import care.data4life.sdk.log.Log
 import care.data4life.sdk.log.Logger
 
-abstract class BaseClient(
```
me-d4l, line 26: Out out pure curiosity: what was the reasoning behind choosing a private-public constructor pair, rather than sticking with the existing constructor and adding `val userID = userService.userID.blockingGet()` inside the class (or in an `init` block) to initialize the property on construction?

me-d4l (approved):
Looks good to me 👍

sdk-core/src/main/java/care/data4life/sdk/BaseClient.kt
bitPogo: Monday Patch - you are totally right for pointing on that, thx.

wmontwe (approved):
👍

me-d4l (approved):
👍

PR #122, opened 2021-05-05
[SDK-574] Change DataResource

bitPogo:
## Description

This refactors the DataResource.

## Motivation and Context

The DataResource can be improved, so we should do that. This is part of [SDK-574](https://gesundheitscloud.atlassian.net/browse/SDK-574) and depends on of #121.

## How is it being implemented?

It makes an addition to the interface of the DataResource. Also `asByteArray` does not act any longer as a alias for `value` - it returns a clone of the wrapped value.

## How Has This Been Tested?

A test set for the DataResource had been added.

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [ ] I have updated the changelog accordingly.
- [x] I have added tests to cover my changes.
- [x] All new and existing tests passed.

PR #123, opened 2021-05-05
[SDK-574] Add ResourceEncryptionService

bitPogo:
## Description

This introduces a new Service - the `ResourceEncryptionService`. Its purpose is move the encryption logic out of the RecordService. Also the DecryptedRecordMapper is not longer needed and was removed.

## Motivation and Context

The RecordService is still too huge. This a step to make it smaller.
The PR is part of [SDK-574](https://gesundheitscloud.atlassian.net/browse/SDK-574) and depends on #122.

As mentioned it adds a new service alongside with its interface. It moves the en/decryption logic out of the RecordService and removes the DecryptedRecordMapper.

## How Has This Been Tested?

A new test set was added to cover the changes. Also all tests of effected classes had been changed accordingly. 

## Types of changes

## Checklist:


sdk-core/src/main/java/care/data4life/sdk/record/RecordContract.kt
```diff
@@ -30,8 +30,7 @@ import io.reactivex.Completable
 import io.reactivex.Single
 import org.threeten.bp.LocalDate
 
-interface RecordContract {
-
+class RecordContract {
```
wmontwe, line 33: Why was this changed to a class?

sdk-core/src/main/java/care/data4life/sdk/network/model/NetworkModelContract.kt
```diff
@@ -20,51 +20,12 @@ import care.data4life.crypto.GCKey
 import care.data4life.sdk.data.DataResource
 import care.data4life.sdk.fhir.Fhir3Resource
 import care.data4life.sdk.fhir.Fhir4Resource
-import care.data4life.sdk.lang.CoreRuntimeException
 import care.data4life.sdk.lang.DataValidationException
 import care.data4life.sdk.tag.Annotations
 import care.data4life.sdk.tag.EncryptedTagsAndAnnotations
 import care.data4life.sdk.tag.Tags
 
 class NetworkModelContract {
```
wmontwe, line 28: I notice that contracts are classes, why that?

bitPogo: EncryptionService and Version are `internal`.

sdk-core/src/main/java/care/data4life/sdk/record/RecordContract.kt
bitPogo: Good catch..this one seemed to slipped while rebasing.

bitPogo: done

sdk-core/src/main/java/care/data4life/sdk/network/model/RecordEncryptionService.kt
```diff
@@ -0,0 +1,331 @@
...
+import care.data4life.sdk.lang.CoreRuntimeException
+import care.data4life.sdk.lang.DataValidationException
+import care.data4life.sdk.model.ModelContract
+import care.data4life.sdk.model.ModelContract.ModelVersion.Companion.CURRENT
+import care.data4life.sdk.tag.Annotations
+import care.data4life.sdk.tag.TaggingContract
+import care.data4life.sdk.tag.Tags
+import care.data4life.sdk.wrapper.WrapperContract
+
+class RecordEncryptionService(
```
alessioborraccino, line 36: is final class a thing in Kotlin?

sdk-core/src/main/java/care/data4life/sdk/network/model/RecordEncryptionService.kt
```diff
@@ -0,0 +1,331 @@
...
+        if (encryptedResource.isEmpty()) {
+            throw CoreRuntimeException.InternalFailure()
+        }
+
+        if (!modelVersion.isModelVersionSupported(version)) {
+            throw DataValidationException.ModelVersionNotSupported("Please update SDK to latest version!")
+        }
+    }
+
+    private fun getCommonKey(commonKeyId: String, userId: String): GCKey {
```
alessioborraccino, line 258: Naming is a bit confusing to me, we have 
`fetchCommonKey()` which returns a pair of GCKey and id
`getCommonKey()` which returns a GCKey 

and i wouldnt know which one refers to which by reading method names

sdk-core/src/main/java/care/data4life/sdk/network/model/RecordEncryptionService.kt
```diff
@@ -0,0 +1,331 @@
...
+        return if (commonKeyStored) {
+            cryptoService.getCommonKeyById(commonKeyId)
+        } else {
+            // TODO: This should be in a different Service
+            val commonKeyResponse = apiService.fetchCommonKey(
+                alias,
+                userId,
+                commonKeyId
+            ).blockingGet()
+
```
alessioborraccino, line 269: I see the TODO :) we created a `CommonKeyService` to wrap up all the CommonKey shenanigans

sdk-core/src/main/java/care/data4life/sdk/network/model/RecordEncryptionService.kt
```diff
@@ -0,0 +1,331 @@
...
+        val commonKey = getCommonKey(encryptedRecord.commonKeyId, userId)
+        val (dataKey, attachmentKey) = decryptKeys(
+            commonKey,
+            encryptedRecord.encryptedDataKey,
+            encryptedRecord.encryptedAttachmentsKey
+        )
+
+        return buildRecord(
+            encryptedRecord.identifier,
+            fhirService.decryptResource(dataKey, tags, encryptedRecord.encryptedBody),
```
alessioborraccino, line 321: is there any way that you can avoid using the `fhirService` here? seems very out of place to me, If this is a record(resource)cryptor

alessioborraccino (changes requested):

sdk-core/src/main/java/care/data4life/sdk/network/model/RecordEncryptionService.kt
bitPogo: Have suggestions for better names?

sdk-core/src/main/java/care/data4life/sdk/network/model/RecordEncryptionService.kt
bitPogo: By default, all classes in Kotlin are final.

sdk-core/src/main/java/care/data4life/sdk/network/model/RecordEncryptionService.kt
bitPogo: Indeed...the whole key stuff should be encapsulated somewhere else...but to do it right now cause more trouble than it helps. However this will happen once Koin (DI) is in place.

sdk-core/src/main/java/care/data4life/sdk/network/model/RecordEncryptionService.kt
bitPogo: No, we cannot avoid that for now. The FhirService is actually more a ResourceCryptoService. We'll come to this once we bubbled up to the corresponding part in the SDK.

sdk-core/src/main/java/care/data4life/sdk/network/model/RecordEncryptionService.kt
bitPogo: On a second look...and since its technically a private method...I guess it maps to the `getCommonKey` method you mentioned. Sure, the signature is different in terms of the parameter but the name and return value matches.

sdk-core/src/main/java/care/data4life/sdk/network/model/RecordEncryptionService.kt
bitPogo: I think the root cause, is more like, then we should slice out this whole mapping record stuff out (same goes for the `fromRecord` method - it will live in the future somewhere else). However I did not it here since it causes much less effort, when finally Koin is in place.

bitPogo: Totally forgot, we'll get rid of the entire mappings of the resources internally, since they will be wrapped. This will in total improve the whole resource handling as well as the readability, but this something which will be seen later in time, since there’s  still much more in queue before.
However I made a follow up to address the overall naming issues - #128.

sdk-core/src/main/java/care/data4life/sdk/network/model/RecordEncryptionService.kt
```diff
@@ -0,0 +1,331 @@
...
+    private val apiService: ApiService,
+    private val taggingService: TaggingContract.Service,
+    private val tagEncryptionService: TaggingContract.EncryptionService,
+    private val guard: NetworkModelContract.LimitGuard,
+    private val cryptoService: CryptoContract.Service,
+    private val fhirService: FhirContract.Service,
+    private val dateTimeFormatter: WrapperContract.DateTimeFormatter,
+    private val modelVersion: ModelContract.ModelVersion
+) : NetworkModelContract.EncryptionService {
+    private fun <T : Fhir3Resource> buildFhir3Record(
```
wmontwe, line 47: RecordEncryptionService is now also responsible for mapping, which blows up the responsibility. Ideally it could delegate that work.

wmontwe (approved):
RecordEncryptionService is too heavy but not a deal breaker.

alessioborraccino (approved):

PR #131, opened 2021-05-10
[SDK-562] Change ApiService to Kotlin

bitPogo:
## Description

This simply converts the ApiService from Kotlin to Java.

## Motivation and Context

All classes needed to be in Kotlin, since the SDK goes into KMP.
This is part of [SDK-562](https://gesundheitscloud.atlassian.net/browse/SDK-562) and depends on #128.

It converts the ApiService to Kotlin.

## How Has This Been Tested?

No additional tests had been added, however the Java module tests stop working and had been ignored. They will replaced with the next PR.

## Types of changes

## Checklist:


sdk-core/src/main/java/care/data4life/sdk/network/ApiService.kt
```diff
@@ -0,0 +1,413 @@
...
+     * @param environment         Deployment environment
+     * @param clientID            Client ID
+     * @param clientSecret        Client secret
+     * @param platform            Usage platform (D4L, S4H)
+     * @param connectivityService Connectivity service
+     * @param clientName          Client name
+     * @param staticAccessToken   Prefetched OAuth token - if not null, it will be used directly (no token renewal).
+     * @param debug               Debug flag
+     */
+    init {
```
wmontwe, line 410: this should be at the top as it is part of the initialization of the class and should follow the general Kotlin Coding Conventions -> see [class layout](https://kotlinlang.org/docs/coding-conventions.html#source-file-organization)

sdk-core/src/main/java/care/data4life/sdk/network/ApiService.kt
```diff
@@ -0,0 +1,413 @@
...
+    private val environment: NetworkingContract.Environment,
+    private val clientID: String,
+    private val clientSecret: String,
+    private val platform: String?,
+    private val connectivityService: NetworkConnectivityService,
+    private val clientName: String?,
+    staticAccessToken: ByteArray?,
+    private val debug: Boolean
+) : NetworkingContract.Service {
+    private var service: IHCService? = null
```
wmontwe, line 63: why is this nullable? couldn't it be initialized in the init() block?

wmontwe (changes requested):

sdk-core/src/main/java/care/data4life/sdk/network/ApiService.kt
bitPogo: You are totally right about this...but I like not put effort into it, since this class had been already rewritten later in the current chain (I will tag the PR once it is uploaded). This step was only necessary to enable proper testing.

sdk-core/src/main/java/care/data4life/sdk/network/ApiService.kt
bitPogo: You are totally right about this...but I like not put effort into it, since this class had been already rewritten later in the current chain (I will tag the PR once it is uploaded). This step was only necessary to enable proper testing.

sdk-core/src/main/java/care/data4life/sdk/network/ApiService.kt
bitPogo: Please see #139.

sdk-core/src/main/java/care/data4life/sdk/network/ApiService.kt
bitPogo: Please see #139.

wmontwe (approved):
👍

PR #132, opened 2021-05-10
[SDK-566] Change ApiService module tests

bitPogo:
## Description

This adds a rewritten version of the ApiService module tests. This is part of [SDK-566](https://gesundheitscloud.atlassian.net/browse/SDK-566) and depends on #131.

## Motivation and Context

Currently the ApiService module tests are not working and being in bad shape, so they needed maintenance.

It adds a fresh written module tests for the ApiService.

## How Has This Been Tested?

This is about tests.

## Types of changes

## Checklist:


sdk-core/src/test/java/care/data4life/sdk/network/ApiServiceModuleTest.kt
```diff
@@ -0,0 +1,725 @@
...
+        assertNull(request.headers[HEADER_ALIAS])
+        assertEquals(
+            actual = request.headers[HEADER_AUTHORIZATION],
+            expected = "Bearer $authToken"
+        )
+        assertEquals(
+            actual = request.method,
+            expected = "DELETE"
+        )
+    }
```
bitPogo, line 596: I just want make obvious, that this should work, but it does not. I even not see, why `deleteDocument` has to be different then `deleteRecord` in terms of returning a Completable. Can we align `deleteDocument` and `deleteRecord` so they working similar?

sdk-core/src/test/java/care/data4life/sdk/network/ApiServiceModuleTest.kt
```diff
@@ -0,0 +1,725 @@
...
+
+    data class TestConnection(
+        override val isConnected: Boolean
+    ) : NetworkConnectivityService
+
+    @Before
+    fun setUp() {
+        server = MockWebServer()
+
+        every { env.getApiBaseURL(any()) } returns "https://www.i-hate-fat-constructors.com"
```
wmontwe, line 74: this should be example.com

wmontwe (changes requested):

sdk-core/src/test/java/care/data4life/sdk/network/ApiServiceModuleTest.kt
bitPogo: done

wmontwe (approved):
👍

PR #134, opened 2021-05-12
Change updateData

bitPogo:
## Description

This makes a very small improvement on the `updateData` method in the RecordService.

## Motivation and Context

`updateData` is one of the methods which needs much further improvements, this is only a small step to make it more understandable.

It use now `getOrElse` instead of a direct null check, when match the oldAttachmentId against the new one.

## How Has This Been Tested?

No new test needed, since no behaviour changed.

## Types of changes

## Checklist:

### DIFF
--- a/sdk-core/src/main/java/care/data4life/sdk/RecordService.kt
+++ b/sdk-core/src/main/java/care/data4life/sdk/RecordService.kt
@@ -965,10 +965,14 @@ class RecordService internal constructor(
                         )
                     newAttachment.id == null -> validAttachments.add(newAttachment)
                     else -> {
-                        val oldAttachment = oldAttachments[newAttachment.id]
-                            ?: throw DataValidationException.IdUsageViolation(
-                                "Valid Attachment.id expected"
-                            )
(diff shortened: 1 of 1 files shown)


sdk-core/src/main/java/care/data4life/sdk/RecordService.kt
```diff
@@ -965,10 +965,14 @@ class RecordService internal constructor(
                         )
                     newAttachment.id == null -> validAttachments.add(newAttachment)
                     else -> {
-                        val oldAttachment = oldAttachments[newAttachment.id]
-                            ?: throw DataValidationException.IdUsageViolation(
-                                "Valid Attachment.id expected"
-                            )
+                        val oldAttachment = oldAttachments.getOrElse(
```
alessioborraccino, line 968: it was somehow easier for me to understand before 😅 
At least now i have it clear either way

wmontwe: I would prefer the previous version as it also fits to how we use it in other places

bitPogo:
I close this in favour for #143.

PR #140, opened 2021-05-12
[SDK-575] Bump OkHTTP

bitPogo:
## Description

This updates OkHTTP from  4.7.2 to 4.9.1

## Motivation and Context

This is part of [SDK-575](https://gesundheitscloud.atlassian.net/browse/SDK-575) and depends on #139.

## How is it being implemented?

Simply bumping the newer version.

## How Has This Been Tested?

No additional tests are needed.

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have updated the changelog accordingly.
- [ ] I have added tests to cover my changes.
- [x] All new and existing tests passed.

PR #141, opened 2021-05-12
[SDK-572] Add JS Legacy Encoding

bitPogo:
## Description

This concludes the actual Tagging compatibly series.

## Motivation and Context

Currently KMP is not able to see legacy JS Records.
This is part of [SDK-572](https://gesundheitscloud.atlassian.net/browse/SDK-572) and depends on #140.

This introduces 2 1/2 completely new classes - the SearchTagBuilder, SearchTags (simple data class) and CompatiblyEncoding.
CompatiblyEncoding follows its names and takes care of the actual 3 different encodings.
SearchTagBuilder is entitled to build the new AND/OR Tag schema in order to make a single API Call.
SearchTags is only a simple Wrapper around the formatted tags.
Alongside the CompatiblyService had been refactored - it now does not longer make the actual API calls. However it is still in charge of controlling the compatibility  Encoding -> Encrypting -> TagFormatting flow.
The RecordService now took back the control of making the actual calls, while resolving the SearchTag via the the CompatibilityService.
Also the Tagging itself got a minor refactoring, in order make this changes more easier to do. It basically slices the former TagCryptoHelper in 2 parts (Encoding and Converter) and wraps the URLEncoder/URLDecoder.

## How Has This Been Tested?

All new classes got unit tests to cover their behaviour. Existing test sets had been altert to match their new behaviour.

## Types of changes

## Checklist:


sdk-core/src/main/java/care/data4life/sdk/migration/RecordCompatibilityService.kt
```diff
@@ -19,138 +19,101 @@ package care.data4life.sdk.migration
...
+            tagGroupKey
         )
     }
 
-    private fun needsDoubleCall(
-        encodedAndEncryptedTags: List<String>,
-        encryptedTags: List<String>
-    ): Boolean = encodedAndEncryptedTags.sorted() != encryptedTags.sorted()
+    private fun addTags(
+        tagGroupKey: String,
```
alessioborraccino, line 51: took me a while to understand this was the `key` part of the key value pair for the tags, i thought it was another encryption key.

sdk-core/src/main/java/care/data4life/sdk/migration/RecordCompatibilityService.kt
```diff
@@ -19,138 +19,101 @@ package care.data4life.sdk.migration
...
+        val builder = searchTagsBuilderBuilderFactory.newBuilder()
+        val tagEncryptionKey = cryptoService.fetchTagEncryptionKey()
 
-        return Single.zip(
-            apiService.getCount(alias, userId, encodedAndEncryptedTags.joinToString(",")),
-            apiService.getCount(alias, userId, encryptedTags.joinToString(",")),
-            BiFunction<Int, Int, Int> { c1, c2 -> c1 + c2 }
-        )
+        mapTags(tags, tagEncryptionKey, builder)
+        mapAnnotations(annotation, tagEncryptionKey, builder)
```
alessioborraccino, line 115: as a matter of taste, i'd like better for the builder to be configured afterwards, meaning that `mapTags` returns tags, `mapAnnotations` returns annotations, and then the builder would add the results later on. 

This way `mapTags` and `mapAnnotations` become more predictable functions without side effects and the builder gets built and sealed in one way.

sdk-core/src/main/java/care/data4life/sdk/migration/RecordCompatibilityService.kt
```diff
@@ -19,138 +19,101 @@ package care.data4life.sdk.migration
...
+    private fun mapAnnotations(
+        annotations: Annotations,
+        tagEncryptionKey: GCKey,
+        builder: NetworkingContract.SearchTagsBuilder
+    ) {
+        annotations.map { annotation ->
+            Pair(
+                ANNOTATION_KEY + DELIMITER,
+                compatibilityEncoder.encode(annotation).copy(
+                    second = compatibilityEncoder.normalize(annotation)
```
alessioborraccino, line 94: why is the second one normalized twice? doesn't encode normalize the second parameter here already?

sdk-core/src/main/java/care/data4life/sdk/network/NetworkingContract.kt
```diff
@@ -117,28 +140,63 @@ interface NetworkingContract {
...
-        const val REQUEST_TIMEOUT = 2
+        const val REQUEST_TIMEOUT: Long = 2
         const val HEADER_ALIAS = "gc_alias"
-        const val HEADER_ACCESS_TOKEN = "access_token"
         const val HEADER_AUTHORIZATION = "Authorization"
-        const val HEADER_BASIC_AUTH = "basic_auth"
-        const val HEADER_GC_SDK_VERSION = "GC-SDK-Version"
+        const val ACCESS_TOKEN_MARKER = "access_token"
+        const val BASIC_AUTH_MARKER = "basic_auth"
+        const val HEADER_GC_SDK_VERSION = "gc-sdk-version"
```
alessioborraccino, line 190: please remember to change this (i guess its done in the other PR)

sdk-core/src/main/java/care/data4life/sdk/network/util/ClientFactory.kt
```diff
@@ -0,0 +1,118 @@
...
+        environment: NetworkingContract.Environment,
+        platform: String,
+    ): OkHttpClient.Builder {
+        return builder
+            .certificatePinner(
+                CertificatePinnerFactory.getInstance(platform, environment)
+            )
+    }
+
+    private fun determineAuthMethod(
```
alessioborraccino, line 42: Since this method is actually adding the right interceptor according to which method it is, this method is kinda doing 2 things. Might be worth it to make a method that returns the right Interceptor according to authMethod and then adding the Interceptor later outside of the function.

I saw thats the pattern of most `determineX` methods, so i'd say same comment applies to the other ones as well

sdk-core/src/main/java/care/data4life/sdk/network/NetworkingContract.kt
```diff
@@ -84,23 +86,44 @@ interface NetworkingContract {
...
+        fun getInstance(platform: String, env: Environment): CertificatePinner
+    }
+
+    internal interface Interceptor : okhttp3.Interceptor {
         override fun intercept(chain: okhttp3.Interceptor.Chain): Response
     }
 
-    interface CertificatePinnerFactory {
-        fun getInstance(baseUrl: String, pin: String): CertificatePinner
+    internal interface PartialInterceptor<T : Any> {
```
alessioborraccino, line 97: More of a comment/question: 

Is there a need for actually creating the concept of a `PartialInterceptor`? 
Might it be worth having only plain interceptors to keep the interfaces simple? In my understanding, it feels more like a namespacing/dependency issue than anything else (as in retryToken and oauthFetchToken both are needed and make sense in the context of Oauth interceptor), but the cost of having a partial interface added a lot overhead on understanding whats happening.

Moreover when does an interceptor become partial? technically all of them could be partial

alessioborraccino (changes requested):

sdk-core/src/main/java/care/data4life/sdk/migration/CompatibilityEncoder.kt
```diff
@@ -0,0 +1,54 @@
...
+    override fun encode(tagValue: String): Triple<String, String, String> {
+        val validEncoding = tagEncoding.encode(tagValue)
+        val normalizedTag = tagEncoding.normalize(tagValue)
+        val jsLegacyEncoding = mapJSExceptions(
+            urlEncoding.encode(normalizedTag)
+        )
+
+        return Triple(
+            validEncoding,
+            normalizedTag, // aka KMPLegacyTags
```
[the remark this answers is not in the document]
bitPogo: it's lowercased and trimmed and is aligned to the TaggingEncoding.

sdk-core/src/main/java/care/data4life/sdk/migration/MigrationContract.kt
```diff
@@ -16,34 +16,34 @@
...
 import care.data4life.sdk.tag.Tags
-import io.reactivex.Observable
-import io.reactivex.Single
 
 @Retention(AnnotationRetention.SOURCE)
 @MustBeDocumented
 annotation class Migration(val message: String)
 
-interface MigrationContract {
+class MigrationContract {
```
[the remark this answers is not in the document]
bitPogo: It's just a vehicle...I needed to change that, since the contract now houses internal interfaces.

sdk-core/src/main/java/care/data4life/sdk/SdkContract.kt
```diff
@@ -269,6 +269,10 @@ interface SdkContract {
         ): Task
     }
 
+    companion object {
+        const val VERSION = "1.10.0"
```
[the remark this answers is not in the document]
bitPogo: This actually belongs to #133. However you are completely right, but I currently do not how to do it for jvm...so my bet was, that @wmontwe has a word about this.

sdk-core/src/main/java/care/data4life/sdk/migration/RecordCompatibilityService.kt
```diff
@@ -19,138 +19,101 @@ package care.data4life.sdk.migration
...
-
-    @Throws(IOException::class)
-    private fun encryptAnnotations(
-        annotations: Annotations,
-        tagEncryptionKey: GCKey
-    ): MutableList<String> {
+    private fun encryptTags(
+        tagEncryptionKey: GCKey,
+        tagGroupKey: String,
+        tagOrGroup: List<String>
```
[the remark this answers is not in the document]
bitPogo: Well, I was not satisfied with `OrComponents` here, since it's technically meant to be a tuple or a set in DNF, which results from the encoding expansion. So I tried to reflect this some how...also this tuple maps to a unique specific key.
This in mind - do you have any suggestions?

sdk-core/src/main/java/care/data4life/sdk/migration/RecordCompatibilityService.kt
```diff
@@ -19,138 +19,101 @@ package care.data4life.sdk.migration
...
         )
     }
 
-    private fun needsDoubleCall(
-        encodedAndEncryptedTags: List<String>,
-        encryptedTags: List<String>
-    ): Boolean = encodedAndEncryptedTags.sorted() != encryptedTags.sorted()
+    private fun addTags(
+        tagGroupKey: String,
+        tagOrGroup: Triple<String, String, String>,
```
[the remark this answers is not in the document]
bitPogo: As I said...its a tuple (which in cludes triples). I think a named tuple would be better, but the overhead this causes barely worth it.

sdk-core/src/main/java/care/data4life/sdk/migration/RecordCompatibilityService.kt
bitPogo: As mentioned above...any suggestions to improve it?

sdk-core/src/main/java/care/data4life/sdk/migration/RecordCompatibilityService.kt
bitPogo: It makes a copy of the existing Triple and replaces the KMP legacy annotations, but they [should not normalized at all](https://github.com/d4l-data4life/hc-sdk-kmp/pull/49/files).

sdk-core/src/main/java/care/data4life/sdk/network/NetworkingContract.kt
bitPogo: it is

sdk-core/src/main/java/care/data4life/sdk/network/util/ClientFactory.kt
bitPogo: The AuthInterceptors are not interchangeable. With a StaticToken we need one Interceptor and without we are need two. Having that in mind, we would need to return a Pair here, where the 2nd item is nullable. And then check the Pair again if the 2n item is actually null.
This way we spare one compare and one null.

bitPogo: Also this belongs into the ClientFactory PR.

sdk-core/src/main/java/care/data4life/sdk/network/NetworkingContract.kt
bitPogo: The Problem which it solves is, that interceptors are basically visitors. We have currently the case with AccessToken, that it does something before the chain proceeds and after the the request returns. And this actions are depend. Since no order of execution is guaranteed, I needed to clue them together in way, that they are still testable.
I do not like it either, but I had no better idea so far...
However if you have...feel free to let me know.

And this belongs to the Interceptor PR.

sdk-core/src/main/java/care/data4life/sdk/migration/RecordCompatibilityService.kt
bitPogo: Well, then I guess I could move it all along on the top level (which I do not wanna do)...however...I can make it shorter.

sdk-core/src/main/java/care/data4life/sdk/migration/MigrationContract.kt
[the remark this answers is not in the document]
alessioborraccino: isnt there another way to namespace interfaces? we have to use enums in swift, not sure whats the used way in kotlin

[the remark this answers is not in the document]
bitPogo: Not with the visibility modifiers...as far as I am aware of.

sdk-core/src/main/java/care/data4life/sdk/migration/RecordCompatibilityService.kt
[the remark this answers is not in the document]
alessioborraccino: to be clear, i guess most of the confusion comes with the fact that the same parameter name `tagOrGroup` is used in two functions, `encryptTags` and `addTags,` but with different type (`List<String>` and `Triple<String,String,String>`) and its hard to go back to these functions and remember what needs to go inside, so maybe unifying will help already

sdk-core/src/main/java/care/data4life/sdk/migration/RecordCompatibilityService.kt
alessioborraccino: labels would help, but also having key/value right after the other for instance.
Also i just notices, isnt tagOrGroup already a list? does it need a .toList() again?

sdk-core/src/main/java/care/data4life/sdk/network/NetworkingContract.kt
alessioborraccino: it's a good solution already. I was just guessing if it was possible to unify them somehow, but i cant think of how right now.

sdk-core/src/main/java/care/data4life/sdk/network/util/ClientFactory.kt
alessioborraccino: so maybe just renaming it would be enough, to something like `addAuthorizationInterceptor`, and inside maybe a helper like `hasStaticAccessToken` for the 2 branches?
Or maybe taking the if one level up directly?

sdk-core/src/main/java/care/data4life/sdk/migration/RecordCompatibilityService.kt
bitPogo: fixed.

sdk-core/src/main/java/care/data4life/sdk/network/util/ClientFactory.kt
bitPogo: I'll make the renaming...`hasStaticAccessToken` would be then only an alias for `staticAccessToken is ByteArray`, since static access and OAuth/Basic are mutual exclusive.

bitPogo: and this is what the method is determining - use static access or OAuth/Basic.

jose-cuadros-data4life (commented):
Good job, I wrote some questions and suggestions, but I didn't find any big issue. In general it would be nice if for future features we could separate them in smaller PRs.

sdk-core/src/main/java/care/data4life/sdk/tag/TagEncoding.kt
```diff
@@ -0,0 +1,47 @@
...
+            throw DataValidationException.AnnotationViolation(
+                "Annotation is empty."
+            )
+        }
+    }
+
+    @Throws(D4LException::class)
+    override fun encode(
+        tag: String
+    ): String = URLEncoding.encode(normalize(tag)).toLowerCase(Locale.US)
```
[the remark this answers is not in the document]
bitPogo: Jap it is, otherwise it takes the default of the System, which is not desirable, since we have no control over that.
Yes, I'll add it to the Contract.

sdk-core/src/main/java/care/data4life/sdk/tag/TaggingContract.kt
```diff
@@ -51,16 +50,22 @@ class TaggingContract {
...
 
-    // TODO: make this package internal
-    interface Helper {
-        fun convertToTagMap(tagList: List<String>): Tags
+    interface Converter {
+        @Throws(D4LException::class)
+        fun toTags(tagList: List<String>): Tags
+    }
 
+    // TODO: make this package internal
```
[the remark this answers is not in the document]
bitPogo: Encoding is intended to be package internal, but since it used somewhere else, it cannot be right now...so I'd like to keep it. Are you okay with that?

sdk-core/src/main/java/care/data4life/sdk/network/util/IHCServiceFactory.kt
```diff
@@ -0,0 +1,48 @@
...
+import care.data4life.sdk.network.IHCService
+import care.data4life.sdk.network.NetworkingContract
+import care.data4life.sdk.network.typeadapter.EncryptedKeyTypeAdapter
+import com.squareup.moshi.Moshi
+import okhttp3.OkHttpClient
+import retrofit2.Retrofit
+import retrofit2.adapter.rxjava2.RxJava2CallAdapterFactory
+import retrofit2.converter.moshi.MoshiConverterFactory
+
+object IHCServiceFactory : NetworkingContract.IHCServiceFactory {
```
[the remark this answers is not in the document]
bitPogo: I can explain the HC, but not the I. Probably @wmontwe knows about the I. 
However I just used the already exiting name in the project, which is only temporary. This whole Service needs to broken down in smaller parts like in JS, then those parts get a better naming.

sdk-core/src/main/java/care/data4life/sdk/migration/RecordCompatibilityService.kt
```diff
@@ -19,138 +19,80 @@ package care.data4life.sdk.migration
...
-        alias: String,
-        userId: String,
+    override fun resolveSearchTags(
         tags: Tags,
-        annotations: Annotations
-    ): Single<Int> {
-        val (encodedAndEncryptedTags, encryptedTags) = encrypt(tags, annotations)
+        annotation: Annotations
+    ): NetworkingContract.SearchTags {
+        val builder = searchTagsBuilderBuilderFactory.newBuilder()
```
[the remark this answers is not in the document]
bitPogo: Glad you spotted it...I renamed the factory a gezillion times so the last time it simply slipped.

sdk-core/src/main/java/care/data4life/sdk/network/ApiService.kt
```diff
@@ -0,0 +1,224 @@
...
+        environment
+    )
+
+    override fun fetchCommonKey(
+        alias: String,
+        userId: String,
+        commonKeyId: String
+    ): Single<CommonKeyResponse> = service.fetchCommonKey(alias, userId, commonKeyId)
+
+    // TODO: Is this method even in use?
```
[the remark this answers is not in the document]
bitPogo: I like to leave it there, since this topic is still not solved....and it belongs to actually an other PR (below in the chain)...

sdk-core/src/main/java/care/data4life/sdk/network/util/IHCServiceFactory.kt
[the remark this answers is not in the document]
jose-cuadros-data4life: Ah okay, so this was done before and was not added now. I see... yeah, I have the same problem, HC I guess what it means, but not the I. Yes, then that can be refactor as extra tickets, no problem 👌

sdk-core/src/main/java/care/data4life/sdk/tag/TaggingContract.kt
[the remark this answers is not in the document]
jose-cuadros-data4life: Yes, it's fine 👍

jose-cuadros-data4life (approved):
👍

bitPogo:
Closed in favour for #145. I will address things, which had been spotted here, but belong to other PRs, when they are on top of the ReviewChain.

PR #146, opened 2021-05-18
[SDK-445] Add SDKConfig

bitPogo:
## Description

This adds the SDKConfig class, which is dynamically generate in order to distribute currently only the version from gradle into the production code.

## Motivation and Context

Currently the version is static, which can be distributed automatically. Also this is part/was requested in the scope of [SDK-445](https://gesundheitscloud.atlassian.net/browse/SDK-445).

It adds to the current build script of the sdk-core a new task which generates the SDKConfig from a template in injects there the version from gradle. The task runs before the compiler. Also `clean` had been modified to been able to purge the SDKConfig from the production code. However the gitignore had not been altered to allow the distribution of the config via this repo.

## How Has This Been Tested?

No additional tests needed since a failure results in a compiler error.

## Types of changes

## Checklist:


alessioborraccino (approved):

wmontwe (commented):
the ingestion client is not using the version?

sdk-ingestion/src/main/java/care/data4life/sdk/Data4LifeClient.java
```diff
@@ -139,7 +138,7 @@ public static Data4LifeClient init(byte[] accessToken,
                 platform,
                 networkConnectivityService,
                 NetworkingContract.Clients.INGESTION,
-                VERSION,
+                SDKConfig.version,
```
bitPogo, line 141: @wmontwe here -> it does make use of it.

bitPogo:
The ingestion client does use the version -> I marked it in the code, so you can easier spot it.

sdk-ingestion/src/main/java/care/data4life/sdk/Data4LifeClient.java
wmontwe: 👍

sdk-core/build.gradle.kts
```diff
@@ -74,3 +74,31 @@ dependencies {
     testImplementation(Dependencies.Java.Test.okHttpMockWebServer)
     testImplementation(Dependencies.Java.Test.jsonAssert)
 }
+
+val templatesPath = "${projectDir}/src/main/resources/templates"
+val configPath = "${projectDir}/src/main/java/care/data4life/sdk/config"
```
wmontwe, line 79: this should go to `src-gen` folder like in the FHIR SDK KMP to avoid accidential changes

sdk-core/src/main/resources/templates/SDKConfig.tmpl
```diff
@@ -0,0 +1,22 @@
...
+ * between you and D4L.
+ *
+ * If you are interested in licensing the SDK for your own applications/third-party
+ * applications and/or if you’d like to contribute to the development of the SDK, please
+ * contact D4L by email to help@data4life.care.
+ */
+
+package care.data4life.sdk.config
+
+// This is generated do not alter it
```
wmontwe, line 19: if moved to `src-gen` this comment is not necessary

bitPogo: Will do

bitPogo: done

sdk-core/build.gradle.kts
bitPogo: done

wmontwe (approved):
👍

PR #164, opened 2021-06-03
[SDK-563] Change AttachmentService tests

bitPogo:
## Description

This makes a rewrite of the current tests for the AttachmentService to align them with the established test pattern.

## Motivation and Context

This is part of [SDK-563](https://gesundheitscloud.atlassian.net/browse/SDK-563).

There is one change which rolls back a untracked behaviour change which was introduced in December and does not align with the previous intended version.

## How Has This Been Tested?

This is mostly about tests.

## Types of changes

## Checklist:

### DIFF
--- a/sdk-core/src/main/java/care/data4life/sdk/attachment/AttachmentService.kt
+++ b/sdk-core/src/main/java/care/data4life/sdk/attachment/AttachmentService.kt
@@ -25,7 +25,6 @@ import care.data4life.sdk.lang.ImageResizeException
 import care.data4life.sdk.log.Log
 import care.data4life.sdk.util.Base64.decode
 import care.data4life.sdk.util.Base64.encodeToString
-import care.data4life.sdk.util.HashUtil.sha1
 import care.data4life.sdk.wrapper.WrapperContract
 import io.reactivex.Observable
 import io.reactivex.Single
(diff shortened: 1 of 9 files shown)


sdk-core/src/main/java/care/data4life/sdk/attachment/AttachmentService.kt
```diff
@@ -125,8 +126,10 @@ class AttachmentService internal constructor(
                     originalData,
                     if (position == POSITION_PREVIEW) DEFAULT_PREVIEW_SIZE_PX else DEFAULT_THUMBNAIL_SIZE_PX
                 )
-                if (downscaledId != null)
-                    additionalIds.add(downscaledId)
+
+                if (downscaledId == null) return null
+
+                additionalIds.add(downscaledId)
```
bitPogo, line 132: Please note: This is the mentioned rollback. See also [here](https://github.com/d4l-data4life/hc-sdk-kmp/blob/aab1e4a026242912c387f1557f60593393503983/sdk-core/src/main/java/care/data4life/sdk/AttachmentService.java#L83-L85) .
However this only a theoretical bug, since if no JEPGWriter is once available it should not be if it try it a 2nd time (except there is a super weird platform in place)

sdk-core/src/main/java/care/data4life/sdk/attachment/AttachmentService.kt
```diff
@@ -36,11 +35,13 @@ class AttachmentService internal constructor(
     // TODO move imageResizer to thumbnail service
     private val imageResizer: AttachmentContract.ImageResizer
 ) : AttachmentContract.Service {
+    private val hasher: AttachmentContract.Hasher = AttachmentHasher
```
wmontwe, line 38: I noticed this in several places, why not adding this to the constructor?

bitPogo: Since, like in other modules, this could be also done by simply reference them, it's more kind of a marker. But you are right, these things should be go away in any place once we make use of Koin (hopefully very soon). However putting them in the constructor now is more of a test concern rather than a structural necessity.

wmontwe (approved):
👍

PR #174, opened 2021-07-20
Bump Update

bitPogo:
This bumps Updates especially for util and FHIR-Helper.

## Description

As stated in the summary it bumps updates. To make them work the setup needed to be changed.

## Motivation and Context

Compatibility with the DD-SDK.

n/a

## How Has This Been Tested?

n/a

## Types of changes

## Checklist:

### DIFF
--- a/crypto-common/src/main/kotlin/care/data4life/crypto/GCKeyPair.kt
+++ b/crypto-common/src/main/kotlin/care/data4life/crypto/GCKeyPair.kt
@@ -15,6 +15,7 @@
  */
 
 package care.data4life.crypto
+import care.data4life.crypto.util.Serializable
 
 expect class GCKeyPair(
     algorithm: GCRSAKeyAlgorithm,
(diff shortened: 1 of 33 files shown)


crypto-common/src/main/kotlin/care/data4life/crypto/GCKeyPair.kt
```diff
@@ -15,8 +15,7 @@
  */
 
 package care.data4life.crypto
-
-import care.data4life.sdk.util.Serializable
+import care.data4life.crypto.util.Serializable
```
alessioborraccino, line 18: why importing this if its not used? or i dont see it?

alessioborraccino (changes requested):

crypto-common/src/main/kotlin/care/data4life/crypto/GCKeyPair.kt
bitPogo: Good question...will have a look, but I simply the old with new signature.

bitPogo: I overlooked it myself - look what GCKeyPair is actually extending

alessioborraccino (approved):

ouafae31 (approved):

PR #178, opened 2021-08-14
[SDK-633] Add iOS legacy tags

bitPogo:
This adds legacy tags for iOS.

## Description

In search/countFhirStu3/r4Records calls, add to the already existing tag variants also the old iOS one: 

Lowercased FIRST, and then percent encoded, leaving the percent encoding uppercase.
(Eg. fhirversion=3.0.1 needs to be encoded into fhirversion=3%2E0%2E1)

## Motivation and Context
To make old iOS documents visible.

It extends the CompatibilityEncoder and adjusts the RecordCompatibilityService accordingly.

## How Has This Been Tested?

A new test for the CompatibilityEncoder had been added and the module tests had been adjusted.

## Types of changes

## Checklist:


CHANGELOG.adoc
```diff
@@ -16,6 +16,8 @@ toc::[]
 
 === Added
 
+* LegacyEncoding for iOS.
```
alessioborraccino, line 19: Maybe this might need to be more descriptive, as more like Fixed and specifying that now sdk is also able tofilter by tags created with old ios apps

sdk-core/src/main/java/care/data4life/sdk/migration/CompatibilityEncoder.kt
```diff
@@ -36,17 +36,16 @@ internal object CompatibilityEncoder : MigrationInternalContract.CompatibilityEn
...
         val kmpLegacyEncoding = tagEncoding.normalize(tagValue)
-        val jsLegacyEncoding = mapJSExceptions(
-            urlEncoding.encode(kmpLegacyEncoding)
-        )
 
-        return Triple(
-            validEncoding,
-            kmpLegacyEncoding, // aka KMPLegacyTags
-            jsLegacyEncoding
+        return MigrationInternalContract.CompatibilityTag(
```
alessioborraccino, line 42: happy you moved on from the triple, wondering about the naming but i dont have a big opinion there. Maybe it would help to specify that is just for search but i wold be ok also to keep it this way.

alessioborraccino (changes requested):

CHANGELOG.adoc
bitPogo: done

sdk-core/src/main/java/care/data4life/sdk/migration/CompatibilityEncoder.kt
bitPogo: done

alessioborraccino (approved):

